PR #42, opened 2023-12-06
fix: Prioritize special queue param defaults over queue env defaults

evanspearman-a:
The Houdini submitter prioritized default values for queue environment parameters incorrectly. A parameter requires a default, so the special case defaults that the submitter uses were never actually applied. This is particularly important for the parameter "RezPackages" because the queue environment typically will not have a good default because it will be different for each DCC. Because the default Queue environment that most customers will be using uses the empty string as a default, the "RezPackages" parameter would never be populated with anything useful unless the user manually specified something.

Switch the logic around to prioritize overriding the default with the special case default if one exists.

The field will automatically be populated with "houdini deadline_cloud_for_houdini" so the user will not need to manually specify the packages.

I created a DeadlineCloud node and observed that it overrode the value while populating the defaults in the UI.

N/A

No, this is a fixing change.


lucaseck (approved):

epmog (approved):
This now matches the other submitter behaviour, thanks Evan!

PR #130, opened 2024-04-25
fix: update SubmitJobProgressDialog to use storage_profile instead of storage_profile_id

gahyusuh:
In `deadline-cloud` version 0.48, the `start_submission` method of the `SubmitJobProgressDialog` class now accepts `storage_profile` instead of `storage_profile_id`. This change is necessary to update the `deadline-cloud` dependency to version 0.48 and maintain compatibility with the latest version.

Update to fetch storage profile and use it instead of `storage_profile_id` when calling the `start_submission` method of an instance of the `SubmitJobProgressDialog`.

Maintains compatibility with the latest version

Ran unit tests and ensured all passed.

No.

No.

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline == 0.47.*",
+    "deadline == 0.48.*",
     "openjd-adaptor-runtime == 0.7.*",
 ]
 
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -643,6 +643,12 @@ def submit_callback(kwargs):
         queue_id = get_setting("defaults.queue_id")
         storage_profile_id = get_setting("settings.storage_profile_id")
 
+        storage_profile = None
+        if storage_profile_id:
+            storage_profile = api.get_storage_profile_for_queue(
+                farm_id, queue_id, storage_profile_id, deadline
+            )
+
         queue = deadline.get_queue(farmId=farm_id, queueId=queue_id)
 
         queue_role_session = api.get_queue_user_boto3_session(
@@ -663,7 +669,7 @@ def submit_callback(kwargs):
         job_progress_dialog.start_submission(
             farm_id,
             queue_id,
-            storage_profile_id,
+            storage_profile,
             job_bundle_dir,
             queue_parameters,
             asset_manager,


chocecil (approved):

jusiskin:
Can we rename the PR title to something that will make more sense to readers of the CHANGELOG? I don't think there is a functional difference to those using the Houdini submitter, so maybe this just becomes:

```
chore(deps): Update deadline requirement from == 0.47.* to == 0.48.*
```

I see there's already a dependabot PR for this (#129) so maybe it makes sense to just push fixing changes to that PR's branch instead.

chocecil:
> Can we rename the PR title to something that will make more sense to readers of the CHANGELOG? I don't think there is a functional difference to those using the Houdini submitter, so maybe this just becomes:
> 
> ```
> chore(deps): Update deadline requirement from == 0.47.* to == 0.48.*
> ```
> 
> I see there's already a dependabot PR for this (#129) so maybe it makes sense to just push fixing changes to that PR's branch instead.

Are there any documented best practices for a case like this? It doesn't seem necessary to push this PR into #129. 
(new to this world)

jusiskin:
> > Can we rename the PR title to something that will make more sense to readers of the CHANGELOG? I don't think there is a functional difference to those using the Houdini submitter, so maybe this just becomes:
> > ```
> > chore(deps): Update deadline requirement from == 0.47.* to == 0.48.*
> > ```
> > I see there's already a dependabot PR for this (#129) so maybe it makes sense to just push fixing changes to that PR's branch instead.
> 
> Are there any documented best practices for a case like this? It doesn't seem necessary to push this PR into #129. (new to this world)

This PR and the dependabot one are duplicates. I proposed migrating to the dependabot one because:

1.  **it already has a conventional PR title.**

    The Deadline Cloud repositories have standardized on [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). This is because we have automation that uses the the merged commit messages to feed our `CHANGELOG.md` and notes for GitHub Releases when we do a release.

    Our repositories use the [`semantic-prs`](https://github.com/Ezard/semantic-prs) GitHub app to enforce that every pull request has a title or a commit that follows specification and we've configured dependabot to create PRs with conventional titles and commit messages.

    I see we do not document this PR requirement in `CONTRIBUTING.md` in this repo, but we should probably add that detail [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/CONTRIBUTING.md#contributing-via-pull-requests)
2.  **dependabot adds release notes for the dependency being bumped**

    This helpful feature allows reviewers to see what changes are being adopted by bumping the dependency and consider how that might impact the functionality. In the case of this dependency, we can see in the dependabot PR that the dependency had a breaking change which we will now need to communicate to consumers of this repository.

Issue #126, opened 2024-04-24
Bug: directories are skipped when determining assets to upload


lucaseck:
### Expected Behaviour

Any external file references that are directories should be included in the Input Directories of the asset manifest and all files within uploaded as part of job submission.

### Current Behaviour

Directories are being skipped when Parse Files is clicked. The Input Directories are never populated, only Input Filenames.

### Reproduction Steps

To reproduce the issue I:
- created a new geometry node in /obj
- inside the new geometry node I created a File Cache node
- set the Base Folder name on the File Cache node to the absolute path to a local folder containing some test files(doesn't matter if they are actual cache files)
- added new mantra and deadline cloud nodes in /out
- clicked Parse Files in the deadline cloud node
- saw no input directories added
  - to be extra sure I added a debug print statement to [where we grab the list of inputs and verified that the linked directory check](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L61) was skipping the directory specified in my File Cache node.

### Code Snippet

All directories are currently skipped when determining inputs [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L61)
```
for parm, ref in hou.fileReferences():
        if parm:
            if parm.node() == rop_node:
                continue
            if ref.startswith(IGNORE_REF_VALUES):
                continue
            if parm.name() in IGNORE_REF_PARMS:
                continue
        if os.path.isdir(ref):
            continue
        input_filenames.add(ref)
```

lucaseck:
Issue #124 should be kept in mind here since the directory existence check can fail as-is if the path isn't fully evaluated and able to be resolved to an actual directory.

Issue #124, opened 2024-04-23
Bug: folders are sometimes being included as input files instead of directories


lucaseck:
### Expected Behaviour

No folders should be added as input files instead of input directories when the asset manifest is generated on submission.

### Current Behaviour

In some cases once the input "filenames" are expanded they result in directory paths instead of actual file paths. These then get added to the asset manifest as an input file not directory which means files can be missed. This looks to be from variable expansion that Houdini only does when calling .eval() which isn't being done until submission time.

### Reproduction Steps

To get the behaviour I:
- created some geometry in /obj
- created a new material texture in /mat
- changed the `Texture Map` path to  `$HIP/`
- added the material to the geometry
- in /out created new mantra and deadline cloud nodes
- clicked parse files in the deadline cloud node
- saw `$HIP/` listed in the Input Files
- pressed submit

In the saved out asset manifest the following folder path was then in the input files not directories:
```
assetReferences:
  inputs:
    directories: []
    filenames:
    - C:/Users/<USER>/Downloads/houdini_scene/
```

### Code Snippet

The fact that the Texture Map path is not expanded when initially loaded seems to bypass the directory check [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L61):
```
for parm, ref in hou.fileReferences():
        if parm:
            if parm.node() == rop_node:
                continue
            if ref.startswith(IGNORE_REF_VALUES):
                continue
            if parm.name() in IGNORE_REF_PARMS:
                continue
        if os.path.isdir(ref):
            continue
        input_filenames.add(ref)
```

Issue #122, opened 2024-04-23
Bug: parse scene files includes possibly unnecessary files


lucaseck:
### Expected Behaviour

Parsing the scene for input files doesn't include unnecessary files or references that aren't actually files

### Current Behaviour

There are cases where files like `voptype.h voplib.h` are included in the Input Filenames after clicking Parse Files. These files don't appear to be needed for rendering and are skipped during submission as missing files.

There are other cases where [`op:` references](https://www.sidefx.com/docs/houdini/io/op_syntax.html) to other nodes are included in the input files and are also evaluated and skipped as missing files during upload.

### Reproduction Steps

To get the header files to be included:
- in /obj I created a new scene with a single light, geometry(cube), camera
- added a default texture material in /mat
- applied the material texture to the geometry node
- in /out created mantra and deadline cloud nodes
- clicked parse files on the deadline cloud node

The parsed files included:
`voptype.h voplib.h`

To get the `op:` references being included:
- I downloaded the `Rocket Launch` sample scene [provided by SideFX](https://www.sidefx.com/contentlibrary/)
- opened the `protonMfail_smoke.hip`
- in /out I added mantra and deadline cloud nodes
- clicked Parse Files in the deadline cloud node

The parsed files include:
`strcat("op:", opfullpath("sky_env_net/sky_env_map"))`

When submitted this evaluates to `- 'op:'` in the asset manifest and the upload is skipped as the file doesn't exist.

### Code Snippet

Some filtering on files to include vs not is already present [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L53)

```
for parm, ref in hou.fileReferences():
        if parm:
            if parm.node() == rop_node:
                continue
            if ref.startswith(IGNORE_REF_VALUES):
                continue
            if parm.name() in IGNORE_REF_PARMS:
                continue
        if os.path.isdir(ref):
            continue
        input_filenames.add(ref)
```

Things like `opdef:", "oplib:", "temp:"` [are already being skipped](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L27), but not `op:` if it should be skipped.

crowecawcaw:
[...]

PR #132, opened 2024-04-26
fix: handle directories, references, and unnecessary files

crowecawcaw:
There were several related bugs with automatically detecting and attaching files:
- https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/124
- https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/126
- https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/122

When detecting if a parameter is a path or file, evaluate the parameter first to resolve the references and attach the expanded path. Also ignore `op:` data references which are [not files](https://www.sidefx.com/docs/houdini/io/op_syntax.html).

Directories are detected. References in file and directory names are handled correctly. Data references are ignored. 

Expanded unit tests. For manual testing, I created a file cache inside a geometry node and pointed the file cache to directories both with `$HIP` as part of the path and without it. I then opened the Deadline Cloud node and clicked "Parse files". The files and directories were correctly detected and attached.

No documentation needed.

No
----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -51,13 +51,12 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
     asset_references.input_filenames.add(_get_hip_file())
 
     for parm, ref in hou.fileReferences():
-        if not parm:
-            continue
-        if parm.node() == rop_node:
-            continue
-        if ref.startswith(IGNORE_REF_VALUES):
-            continue
-        if parm.name() in IGNORE_REF_PARMS:
+        if (
+            (not parm)
+            or (parm.node() == rop_node)
+            or (ref.startswith(IGNORE_REF_VALUES))
+            or (parm.name() in IGNORE_REF_PARMS)
+        ):
             continue
 
         path = parm.evalAsString()
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -24,7 +24,7 @@ from ._version import version
 
 import hou
 
-IGNORE_REF_VALUES = ("opdef:", "oplib:", "temp:")
+IGNORE_REF_VALUES = ("opdef:", "oplib:", "temp:", "op:")
 IGNORE_REF_PARMS = (
     "taskgraphfile",
     "pdg_workingdir",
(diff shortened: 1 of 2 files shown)


test/deadline_submitter_for_houdini/unit/test_assets.py
```diff
@@ -13,17 +14,38 @@ def test_get_scene_asset_references():
...
+        (temp_parm, "temp:$OS.rat"),
     )
-    a = _get_scene_asset_references(node)
-    assert a.input_filenames == {"$HIP/houdini19.5/otls/Deadline-Cloud.hda", "/some/path/test.hip"}
-    assert a.input_directories == set()
+    mock_os = mock.Mock()
+    mock_os.path.isdir = lambda path: path.endswith("/")
+    mock_os.path.isfile = lambda path: not path.endswith("/")
+
+    print(mock_os.path.isdir("a"))
```
lucaseck: Are these just debug statements left behind?

test/deadline_submitter_for_houdini/unit/test_assets.py
```diff
@@ -13,17 +14,38 @@ def test_get_scene_asset_references():
...
+    file_parm = mock.Mock()
+    file_parm.node.return_value = None
+    file_parm.evalAsString.return_value = "/path/asset.png"
+
     hou.fileReferences.return_value = (
-        (None, "$HIP/houdini19.5/otls/Deadline-Cloud.hda"),
-        (parm, "temp:$OS.rat"),
+        (dir_parm, "$HIP/houdini19.5/"),
+        (file_parm, "$HIP/houdini19.5/otls/Deadline-Cloud.hda"),
+        (temp_parm, "temp:$OS.rat"),
```
lucaseck: Can some additional cases be added or parameterized so that we have `opdef:` `op:` and `oplib:` being skipped captured in the tests.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -24,7 +24,7 @@
 
 import hou
 
-IGNORE_REF_VALUES = ("opdef:", "oplib:", "temp:")
+IGNORE_REF_VALUES = ("opdef:", "oplib:", "temp:", "op:")
```
lucaseck, line 27: Did you do any additional digging or find more references that `op:` always points to another node and never somehow evaluates to a real path? Just want to make sure we aren't missing anything by adding this to the skip list.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -48,19 +48,24 @@ def _get_houdini_version() -> str:
...
     for parm, ref in hou.fileReferences():
-        if parm:
-            if parm.node() == rop_node:
-                continue
-            if ref.startswith(IGNORE_REF_VALUES):
-                continue
-            if parm.name() in IGNORE_REF_PARMS:
-                continue
-        if os.path.isdir(ref):
+        if not parm:
```
lucaseck: nit: Can this be cleaned up a bit either through conjoining some of the conditions or changing to a if->elif->else so it's not so much of a wall of if->continues. If that makes things less readable though we can also leave it as-is.

lucaseck (commented):
Thank you for the contributions. I've left some minor comments.

As well, for #122 this won't fully fix it since I don't see anything about those header files being ignored. If that's going to be a separate PR no worries. Just didn't want to have the issue be closed without all parts being fixed.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
crowecawcaw: I found [this page](https://www.sidefx.com/docs/houdini/io/op_syntax.html) which sounds like `op:` references are always internal scene references and not files. I'll double check with an SME.

test/deadline_submitter_for_houdini/unit/test_assets.py
crowecawcaw: Will do

test/deadline_submitter_for_houdini/unit/test_assets.py
crowecawcaw: Yup, missed it! Will remove

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
crowecawcaw: I'll take a look.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
lucaseck: Confirmed offline that we should be covered here and find to skip `op:`

lucaseck (approved):

gahyusuh (approved):

PR #135, opened 2024-04-29
fix: Windows pathmapping rules

lucaseck:
It was noticed that absolute Windows paths weren't getting mapped at all on the workers. This is because the pathmapping rules were being changed to using double backslashes `\\` as part of a dictionary to string conversion which when set in HOUDINI_PATHMAP Houdini converts them internally into double forward slashes `//`. This meant that the paths would not be mapped as Houdini uses a single forward slash `/` as the separator when selecting files and folders on Windows.
All rules now have a find/replace done on them for `\\`->`/`
Windows absolute paths are now mapped correctly by the adaptor
- Created a scene which referenced an absolute texture path and had an absolute path on the output of the Mantra node
- Ran a submission from a Windows machine to a Linux worker with the current adaptor and saw that the texture was not mapped and the output not found since it was also not mapped
- Tested again after these changes and saw that the texture was mapped correctly and the output file was found and uploaded as output correctly.
No
No
----


src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
```diff
@@ -373,7 +374,19 @@ def _get_houdini_pathmap(self) -> str:
         path_mapping_rules: dict[str, str] = {}
 
         for rule in self._path_mapping_rules:
-            path_mapping_rules[rule.source_path] = rule.destination_path
+            source_path = rule.source_path
+            destination_path = rule.destination_path
+            rule_dict = rule.to_dict()
+
+            # Convert Windows paths to Posix as HOUDINI_PATHMAP turns \\ in
```
mwiebe: I think instead of conditionally calling `as_posix()` when it's Windows, you can just call it regardless of OS. From the Python docs, it indicates that the function is a member of the parent `PurePath`. That would simplify the code to something like:

```python
# Houdini uses forward slashes on all OS
source_path = source_path.as_posix()
destination_path = destination_path.as_posix()
```

lucaseck: I had tried that, but it doesn't seem to work Windows->Linux. I believe what's happening is when the rules are loaded on Linux it defaults to a PurePosixPath underneath which treats the `\` as valid path characters and not a separator to swap, so nothing is changed in the path.

The [implementation](https://github.com/python/cpython/blob/3.12/Lib/pathlib.py#L454) of `as_posix` does a `.replace(sep, "/")` where `sep` is the value of `os.path.sep` which on Linux is already `/` so nothing is being replaced.
For example, running on Linux:
```
>>> from pathlib import PurePath
>>> test_path = "C:\\folder1\\folder2\\output.png"
>>> p_test_path = PurePath(test_path)
>>> p_test_path
PurePosixPath('C:\\folder1\\folder2\\output.png')
>>> p_test_path.as_posix()
'C:\\folder1\\folder2\\output.png'
>>> 
```

epmog (dismissed):

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
mwiebe: Interesting, then I think doing a `.replace(os.path.sep, "/")` would be nicer than the `if/else` on the OS type. That reduces the dependencies on details in the adaptor library, so helps us refactor it.

mwiebe (approved):

epmog (approved):

PR #137, opened 2024-05-01
fix: verify Deadline Cloud render node input exists before submitting

joel-wong-aws:
If the Deadline Cloud ROP does not have an input ROP, then submission of a job will first hash and upload all the files for the job, then attempt to submit the job, then get a confusing error message related to missing steps.

The error is caused by missing an input ROP, but the the message is confusing. Added verification that there is an input ROP at the start of the submission.

Easier onboarding for customers

1. Created a simple render with spheres, light, and camera
2. Created a Mantra ROP
3. Created a Deadline Cloud ROP, but did not connect it to the Mantra ROP
4. Click "Submit" in the Deadline ROP

Before:
image

After:
image

No, not necessary

No

----


lucaseck (approved):
Thanks Joel!

epmog (approved):

PR #144, opened 2024-05-06
fix: output directory detection

### LINKED ISSUE #138: Bug: comp outputs are missed by parse files
### Expected Behaviour

If there is a comp node in the `/out` network and `Parse Files` is clicked the bottom directory of the output path will be added to the detected list of `output_directories`

### Current Behaviour

If there is a comp node in the `/out` network and `Parse Files` is clicked the output from the comp node will not be added to the list of detected `ouput_directories`.

### Reproduction Steps

- Create a new scene in Houdini
- Go to the /out network
- Add a comp node and a deadline-cloud node and connect them
- Click `Parse Files`
- No output directory is added

### Code Snippet

The list of mappings that controls what is detected as a node that can have a detected output_directory is [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L68), any other nodes that should be included need to be added there:
```
rop_dir_map = {
        # Mantra/karma
        "Driver/ifd": "vm_picture",
        "Driver/karma": "picture",
        # Husk
        "Lop/usdrender_rop": _husk_outputs,
        # Arnold
        "Driver/arnold": "ar_picture",
        # Vray
        "Driver/vray_renderer": "SettingsOutput_img_file_path",
        "Sop/rop_vrayproxy": "filepath",
        "Driver/rop_vrayproxy": "filepath",
        # Renderman
        "Driver/ris::3.0": _renderman_outputs,
        # Redshift
        "Driver/Redshift_ROP": "RS_outputFileNamePrefix",
        # Geo
        "Driver/geometry": "sopoutput",
        "Driver/alembic": "filename",
        "Sop/filecache": "file",
    }
```

lucaseck:
Fixes #138 
Many types of nodes were not being included in the output directory detection when using parse files.
- Updated the mappings list for which nodes are detected as having output paths
- Refactored the asset management related functionality out into a new `assets.py` file.
Hopefully a majority of cases should now be covered by the output directory detection
Created a scene file with all of the newly added node types within. Clicked `Parse Files` and verified that the directories specified in each nodes output paths were included. For testing the Fetch and Wedge nodes I made longer chains with valid output nodes at the end of the chains to ensure they were recursively being checked until finding the end node.
No
No
----


src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -65,40 +65,74 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
...
+        "Driver/channel": "chopoutput",  # Channel
+        "Driver/comp": "copoutput",  # Composite
+        "Driver/dop": "dopoutput",  # Dynamics
+        "Driver/fetch": find_output_dirs,  # Fetch
+        "Sop/filecache": "file",  # File Cache
+        "Sop/filecache::2.0": "file",  # File Cache v2
+        "Driver/filmboxfbx": "sopoutput",  # Filmbox FBX
+        "Driver/geometry": "sopoutput",  # Geometry
+        "Lop/usdrender_rop": _husk_outputs,  # Husk
+        "Driver/karma": "picture",  # Karma
```
joel-wong-aws: I believe Karma's node type is `Driver/usdrender`

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -65,40 +65,74 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
...
+    def find_output_dirs(node: hou.Node) -> set[str]:
+        """
+        Given a node, find and return the directories of the output path parms.
+        """
+        out_dirs = set()
         type_name = node.type().nameWithCategory()
         out_parm = rop_dir_map.get(type_name, None)
-        if out_parm is not None:
+
+        if out_parm:
```
joel-wong-aws: Instead of using a nested `if` statement here, can we use guard clauses? That makes it easier to see the return values for special cases e.g. if `out_parm` is not defined

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -65,40 +65,74 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
...
+    """
+    Find and return the set of all output directories detected from the output
+    paths within the nodes leading into the Deadline Cloud ROP node
+    """
+
+    def find_output_dirs(node: hou.Node) -> set[str]:
+        """
+        Given a node, find and return the directories of the output path parms.
+        """
+        out_dirs = set()
```
epmog: Since the outer function isn't called often this should be fine, but just a heads up that the inner function is redefined every time the outer is called.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -65,40 +65,74 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
...
-        if out_parm is not None:
+
+        if out_parm:
             if callable(out_parm):
-                computed_dirs = out_parm(node)
-                asset_references.output_directories.update(computed_dirs)
+                if type_name == "Driver/wedge":
+                    node = node.node(node.parm("driver").eval())
+                elif type_name == "Driver/fetch":
+                    node = node.node(node.parm("source").eval())
```
joel-wong-aws: Would it be better to put this logic into a separate function, like we do for husk and Renderman? That could make the logic more modular

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -65,40 +65,74 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
...
+        """
+        Given a node, find and return the directories of the output path parms.
+        """
+        out_dirs = set()
         type_name = node.type().nameWithCategory()
         out_parm = rop_dir_map.get(type_name, None)
-        if out_parm is not None:
+
+        if out_parm:
             if callable(out_parm):
```
epmog: Edit: this echoes Joel's comment regarding "guard clauses"

alternative: you could also write this with inverted if-statements to de-indent and keep the relevant code to the conditions. Up to you though since this function is currently pretty small and it's not too hard to read.

ie.
```python
if not out_parm:
    return

if not callable(out_parm):
    path = node.parm(out_parm).eval()
    out_dirs.add(os.path.dirname(path))
    return out_dirs

...
```

epmog (dismissed):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
epmog: It may depend on if you're grabbing output vs. other settings like verbosity? Some things will be on the Karma node, and others may be in the inner usdrender node. @lucaseck can confirm when they see this

lucaseck: Yup, can confirm that just adding the Karma node it appears as `Driver/karma` when querying the node using `.type().nameWithCategory()`

lucaseck: Good point though that USD Render is missing. I'll add that one in.

epmog: To the original point though. The output information **_is_** located on `Driver/karma`? ie. it's correct to have both?

lucaseck: Yes, the output information is on `Driver/karma` when using the Karma node.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
lucaseck: I can take another look. It gets a little harder with Wedge and Fetch nodes as we need to recursively go through the chain of nodes they can point to until we hit one that is a regular output node. You can chain multiple Fetch and Wedge nodes together so it's not as simple as Husk and Renderman which are only looking at one level deep I believe.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
lucaseck: Ended up leaving out adding in USD Render for now. It needs more digging into how output paths work within it.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
lucaseck: Good callout, changed in the latest revision.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
lucaseck: Refactored out into a new file for asset related functionality.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
```diff
@@ -0,0 +1,175 @@
...
+
+    if not callable(out_parm):
+        path = node.parm(out_parm).eval()
+        out_dirs.add(os.path.dirname(path))
+        return out_dirs
+
+    detected_dirs = out_parm(node)
+    out_dirs.update(detected_dirs)
+
+    return out_dirs
```
joel-wong-aws: supernit: we don't ever use out_dirs twice for a single case here, so it's not necessary to defined `out_dirs: set[str] = set()`. Instead, we could just do
```
if not out_parm:
    return set()

if not callable(out_parm):
    path = node.parm(out_parm).eval()
    return {os.path.dirname(path))}

return out_parm(node)
```

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
```diff
@@ -0,0 +1,175 @@
...
+
+        path = parm.evalAsString()
+        if os.path.isdir(path):
+            asset_references.input_directories.add(path)
+        if os.path.isfile(path):
+            asset_references.input_filenames.add(path)
+
+        all_inputs = rop_node.inputAncestors()
+        for node in all_inputs:
+            asset_references.output_directories.update(_get_output_directories(node))
```
joel-wong-aws: Should this be unindented? It looks like that was the original implementation and there isn't any need for us to do this step once for each file reference

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
```diff
@@ -0,0 +1,175 @@
...
+        for child in products.GetChildren():
+            if child.GetTypeName() == "RenderProduct":
+                product_name_attr = child.GetAttribute("productName")
+                path = product_name_attr.Get(0)
+                output_directories.add(os.path.dirname(path))
+    return output_directories
+
+
+_ROP_DIR_MAP = {
+    "Driver/alembic": "filename",  # Alembic
```
joel-wong-aws: nit: would be nice to have these in alphabetical order

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
```diff
@@ -0,0 +1,175 @@
...
+            return output_directories
+        for child in products.GetChildren():
+            if child.GetTypeName() == "RenderProduct":
+                product_name_attr = child.GetAttribute("productName")
+                path = product_name_attr.Get(0)
+                output_directories.add(os.path.dirname(path))
+    return output_directories
+
+
+_ROP_DIR_MAP = {
```
joel-wong-aws: nit: are these all ROPs? Some of them start with "sop" or "lop"

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
```diff
@@ -0,0 +1,175 @@
...
+        out_dirs.add(os.path.dirname(path))
+        return out_dirs
+
+    detected_dirs = out_parm(node)
+    out_dirs.update(detected_dirs)
+
+    return out_dirs
+
+
+def _fetch_outputs(node: hou.Node) -> set[str]:
```
joel-wong-aws: Do we need unit tests for these new functions?

lucaseck: Yes, but there are no existing ones for the other functions. I'm planning to take that on next to try and improve the unit tests around the asset stuff.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
lucaseck: They are alphabetical based on the comments on the right which are the actual Node names in Houdini

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
lucaseck: Indented is correct in this case. If the loop over the inputAncestors is moved into `_get_output_directories` then for the Fetch and Wedge cases where there is recursion we will be looping over the input ancestors of those nodes which is not what is intended.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -20,87 +20,16 @@
 from deadline.job_attachments.models import JobAttachmentS3Settings
 
 from .queue_parameters import update_queue_parameters, get_queue_parameter_values_as_openjd
+from .assets import _get_hip_file, _get_asset_references, _parse_files
 from ._version import version
 
 import hou
 
-IGNORE_REF_VALUES = ("opdef:", "oplib:", "temp:", "op:")
-IGNORE_REF_PARMS = (
```
epmog, line 28: Gah, unfortunately we previously marked these as public in a public module. To prevent a needless breaking change let's just import these from the assets module and ignore them for now.

The longer term fix to prevent this type of stuff from coming up is clarifying our public interfaces on the integrations since they're applications. ie. Is it just how we launch the application?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/assets.py
lucaseck: Spoke with Joel offline. We are both talking about different spots and he is correct that this should be unindented one level. Was working and passing tests still, but calling `_get_output_directories` way more times than necessary.

joel-wong-aws (approved):

epmog (approved):

Issue #117, opened 2024-04-17
Bug: login callback does not validate farm or queue values before calling Deadline APIs


AWS-Samuel:
### Expected Behaviour

`get_farm` or `get_queue` is called with non-empty values. 

### Current Behaviour

`get_farm` or `get_queue` is called with empty values on first load leading to undefined behavior. 

### Reproduction Steps

Remove all cached deadline settings,  then login to the submitter. 

### Code Snippet

https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L701-L707:

```
def login_callback(kwargs):
    node = kwargs["node"]
    node.parm("farm").set("<refreshing>")
    node.parm("queue").set("<refreshing>")
    DeadlineLoginDialog.login(parent=hou.qt.mainWindow())
    deadline = api.get_boto3_client("deadline")
    farm_id = get_setting("defaults.farm_id")
    farm_response = deadline.get_farm(farmId=farm_id)
    node.parm("farm").set(farm_response["displayName"])
    queue_id = get_setting("defaults.queue_id")
    queue_response = deadline.get_queue(farmId=farm_id, queueId=queue_id)
    node.parm("queue").set(queue_response["displayName"])
    update_queue_parameters(farm_id, queue_id, node)
```

AWS-Samuel:
If using credentials without access to `deadline:ListFarms` this can show up as the following error:

```
botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the GetFarm operation: User: arn:aws:sts::<account-id>:<role> is not authorized to perform: deadline:ListFarms on resource: arn:aws:deadline:<region>:<account-id>:farm/*
```

A workaround is to set the `farm-id` and `queue-id`  outside of the houdini submitter:

```
deadline config gui
```

PR #148, opened 2024-05-08
fix: improve error messages when farm or queue ID not defined

joel-wong-aws:
When first submitting a job to AWS Deadline Cloud using its render node, I received unexpected error messages with an Access Denied error. These were related to missing default farms and queues.

image

Added a check whether the farm and queue are defined before calling GetQueue in the submitter.

image

image

Easier onboarding for new customers

1. Created a simple scene with geometry, light, camera
2. Connected a Karma ROP to the Deadline Cloud ROP
3. Clicked submit (without adding settings, or after clearing the default farm and queue with `deadline config set defaults.farm_id ""` and `deadline config set defaults.queue_id ""`)

No, not required

No

----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -528,13 +528,29 @@ def submit_callback(kwargs):
                 "houdini-version": _get_houdini_version(),
             }
         )
+        farm_id = get_setting("defaults.farm_id")
+        if not farm_id:
+            hou.ui.displayMessage(
+                "Please configure the farm ID in the AWS Deadline Cloud render node (ROP) settings",
+                title="Farm ID Required",
+                severity=hou.severityType.Warning,
+            )
+            return
+
+        queue_id = get_setting("defaults.queue_id")
+        if not queue_id:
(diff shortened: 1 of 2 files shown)


crowecawcaw (approved):

AWS-Samuel (commented):
Thanks for fixing this bug! The changes here look good, but it looks like we could add the same logic to 
- `settings_callback`
- `login_callback`
- `update_queue_parameters_callback`

in `src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py` 

Which would allow us to resolve https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/117

Do you mind adding the checks for unset farm/queue there too?

AWS-Samuel:
> Thanks for fixing this bug! The changes here look good, but it looks like we could add the same logic to
> 
> * `settings_callback`
> * `login_callback`
> * `update_queue_parameters_callback`
> 
> in `src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py`
> 
> Which would allow us to resolve #117
> 
> Do you mind adding the checks for unset farm/queue there too?

After syncing offline, we've agreed to tackle these other places in a future PR

Ship it :shipit:

AWS-Samuel (approved):

PR #149, opened 2024-05-14
feat!: run simulations sequentially

crowecawcaw:
The submitter previously always rendered each frame as a task, and tasks would be distributed to different Deadline Cloud workers. But some node types like simulations must be run sequentially - each frame depends on the previous one.

Introduce a new sequential rendering strategy where all frames are rendered inside a single Deadline Cloud task which ensures they are sequentially on a single worker.

Specific changes include:
- adding logic to detect simulation nodes
- updating the runData schema to support sending frame ranges instead of single frames

Simulations are run efficiently.

I submitted a hip file with simulations and regular rendering. The simulations resulted in single tasks and the renders resulted in 1 task per frame.

No, but it needs to be.

No. Changing the runData schema could have been a breaking change, but I took care to ensure the old `frame` parameter is still handled correctly.

----


src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -332,18 +370,19 @@ def _get_job_template(rop: hou.Node) -> dict[str, Any]:
...
+
         if "dependency_names" in node:
             deps = [{"dependsOn": d} for d in node["dependency_names"]]
             step["dependencies"] = deps
         steps.append(step)
     job_template = {
         "specificationVersion": "jobtemplate-2023-09",
-        "name": job_name,
+        "name": rop.parm("name").evalAsString(),
+        "description": rop.parm("description").evalAsString() or None,
```
epmog: Does this end up with `null` in the template? Is the behaviour different in the service if a description is `null` vs not there vs an empty string?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -274,44 +291,65 @@ def _get_job_template(rop: hou.Node) -> dict[str, Any]:
...
+                    {"name": "Frame", "range": range_expression, "type": "INT"}
+                ]
+            }
+            task_data_dict["frame_range"] = {
+                "start": "{{Task.Param.Frame}}",
+                "end": "{{Task.Param.Frame}}",
+                "step": 1,
+            }
+
+        task_data = yaml.dump(task_data_dict, default_flow_style=False)
```
epmog: Does this end up changing what the exported templates look like? Based on the test it appears to be the same. Hmm... the readability hasn't been good there in the past. We should be thinking about turning this into a multiline yaml string, but that's future change and there's other things needed to make these generated templates better so just ignore this for now.

Also I haven't used pyyaml all too much, should we also be doing `safe_dump` here?

epmog:
Regardless of if openjd-adaptor-runtime keeps the interface version, we can delay the decision by bumping it regardless.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
crowecawcaw: It does add `null`s. They service accepts them. They are treated the same as if the key were not present. Empty strings are rejected though.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
crowecawcaw: This change does not affect the export templates except for reordering the fields. It produces the same format. I made this change to improve readability and to handle special characters (e.g. `"`) correctly.

Looking at the docs, I don't think we need `safe_dump`, but it produces the same result so I'll swap it out.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
epmog: Sounds good, just wanted to double check. Thanks!

epmog (dismissed):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
epmog: Yea, there's 2 different "readabilities" at play here:
1. The code here
2. The generated template that a user will look at if they want to customize/debug

I was mostly referring to the second since you've greatly improved the first

I was on the fence about that based on the inputs we accept from the user. Thanks!

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
joel-wong-aws: Following up to Morgan's comment: can we skip this argument if there is no description? i.e. rather than passing `"description": None`, why not just not pass the argument? It's not clear if OpenJD _should_ accept a `null` description.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -274,44 +291,65 @@ def _get_job_template(rop: hou.Node) -> dict[str, Any]:
...
+        task_data = yaml.safe_dump(task_data_dict, default_flow_style=False)
+        # Remove single quotes around the frame parameter so it gets interpreted as a int and not a string
+        task_data = task_data.replace("'{{Task.Param.Frame}}'", "{{Task.Param.Frame}}")
+
         step = {
             "name": node["name"],
-            "parameterSpace": {
-                "taskParameterDefinitions": [{"name": "Frame", "range": frame_range, "type": "INT"}]
-            },
+            "parameterSpace": parameter_space,
```
joel-wong-aws: If `parameterSpace` is None, should we exclude this argument?

test/deadline_submitter_for_houdini/unit/test_job_template.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@patch(
+    "deadline.houdini_submitter.python.deadline_cloud_for_houdini.submitter._get_houdini_version",
+    Mock(return_value="19.5.435"),
+)
+@patch(
+    "deadline.houdini_submitter.python.deadline_cloud_for_houdini.submitter._get_hip_file",
+    Mock(return_value="/path/to/hip.hip"),
+)
```
joel-wong-aws: nit: these mocks are repeated multiple times in this file, can we use a fixture instead?

test/deadline_submitter_for_houdini/unit/test_job_template.py
```diff
@@ -0,0 +1,187 @@
...
+            "rop": "/mantra",
+            "wedgenum": "",
+            "wedge_node": "",
+            "start": 1,
+            "end": 5,
+            "step": 1,
+            "render_strategy": RenderStrategy.PARALLEL,
+        }
+    ]
+    template = _get_job_template(mock_node)
```
joel-wong-aws, line 74: nit: there are a lot of tests of private functions here, can we test public ones instead?

crowecawcaw: I think the functions organize and encapsulate different bits of logic reasonably well, and the tests run them through the various cases we care about. Is the private vs public distinction the main thing you're looking at? What do you think about just making the functions public?

crowecawcaw: I looked more at making these public functions, but I think they made more sense as private functions since they're only used within this file. Another pattern we could do is move them into a helpers/utils file and make them public, but that would just spread out the code and IMO would make it less readable. I'm going to leave them as private for now but am open to reorganizing it if you have specific ideas.

joel-wong-aws: Yeah, it's generally an anti-pattern to test (and mock) private functions as it usually means that we are testing implementation details rather than behaviour. Sometimes it is reasonable to test private functions given constraints/time/code structure, but it could also indicate opportunities to refactor the code for better readability, SOLID principles, etc. In this case, I don't think it's a blocker but could be an opportunity for future refactors 🙂

epmog: Something to keep in mind here is that the public interface for applications is different than the interface for libraries and we're not reflecting that correctly in some of these repos.

Effectively, the only public thing in submitters should be the part that actually hooks up the submitter with the DCC, as that's the only thing we're really making available to customers. Everything else is effectively internal/private to the application, but we've done this at the individual function level rather than the module level.

But breaking changes is around the behaviour presented to users and not just the public interface (though we are changing the adaptor public interface).

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -188,6 +199,33 @@ def _get_rop_steps(rop: hou.Node):
...
+        render_strategy = RenderStrategy.SEQUENTIAL
+
+    if node.parm("deadline_cloud_render_strategy"):
+        strategy_string = node.parm("deadline_cloud_render_strategy").evalAsString()
+        if strategy_string.upper() == RenderStrategy.SEQUENTIAL.value:
+            render_strategy = RenderStrategy.SEQUENTIAL
+        elif strategy_string.upper() == RenderStrategy.PARALLEL.value:
+            render_strategy = RenderStrategy.PARALLEL
+        elif strategy_string == "":
+            # Use the submitter's default choice
```
epmog: Any reason to not make this an error? Feels weird to provide another avenue to just do what the submitter picks.

If we remove this case, then we could handle overrides first instead of wasting time and checking what we think the node should be. Could look something like this:

```python
if node.parm("deadline_cloud_render_strategy"):
    strategy_string = node.parm("deadline_cloud_render_strategy").evalAsString()
    if strategy_string.upper() == RenderStrategy.SEQUENTIAL.value:
        return RenderStrategy.SEQUENTIAL
    elif strategy_string.upper() == RenderStrategy.PARALLEL.value:
        return RenderStrategy.PARALLEL
    else:
        raise ValueError(...)

# no overrides, try to figure it out.
if (
    node.type().nameWithCategory() == "Driver/geometry"
    and node.parm("initsim")
    and node.parm("initsim").eval()
):
    return RenderStrategy.SEQUENTIAL

return RenderStrategy.PARALLEL
```

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -188,6 +199,33 @@ def _get_rop_steps(rop: hou.Node):
...
+            pass
+        else:
+            raise Exception(
+                f'node {node.path()} has an unexpected value "{strategy_string}" for its "deadline_cloud_render_strategy" parameter. Choose either {RenderStrategy.PARALLEL.value} or {RenderStrategy.SEQUENTIAL.value}'
+            )
```
epmog: Have we tested how this error shows up to the user to ensure it matches the usability we want?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -188,6 +199,33 @@ def _get_rop_steps(rop: hou.Node):
     return rop_steps
 
 
+# Any changes to this function should be reflected in the user guide.
```
epmog: nit: can move this to a doc-string and point to the file to update

docs/user-guide.md
```diff
@@ -0,0 +1,28 @@
+# User guide
+
+## Parallel vs. Sequential Rendering
+
+For many types of nodes, frames can be rendered independently and in any order. For others like simulations, each frame depends on the result of the previous frame and must be rendered sequentially. The submitter guesses the optimal rendering strategy for each node, but also allows you to override the default.
```
lucaseck: nit: "guesses" just sounds like there's no strategy to it. Maybe `makes assumptions about the optimal...`

docs/user-guide.md
```diff
@@ -0,0 +1,28 @@
...
+By default, if a node is a geometry node with the "Initialize Simulation OPs" option enabled, it will render sequentially. Otherwise the node will render in parallel.
+
+### Overriding the render strategy
+
+You can override the render strategy be creating a `deadline_cloud_render_strategy` parameter with a value of either `SEQUENTIAL` or `PARALLEL`.
+
+To add a parameter to a node:
+1. Right click a node in the `out` context, select "Parameters and Channels" then click "Edit Paramter Interface".
+2. Under "Create Parameteres", select "Ordered Menu" and click the right arrow to add it to the "Existing Parameters" list.
+3. Select the new parameter in under "Existing Paramters", then edit its configuration under "Parameter Description":
```
lucaseck: `in under` -> `under`

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
crowecawcaw: It pops up an error message that looks like this: Screenshot 2024-05-22 at 7 31 34 AM

docs/user-guide.md
```diff
@@ -0,0 +1,28 @@
...
+    a. Under the "Parameter" tab, set "Name" to `deadline_cloud_render_strategy` and "Label" to "Deadline Cloud Render Strategy"
```
lucaseck, line 20: Formatting is off here in the rendered document. Should be a. and b. I believe and some newlines missing

scripts/build_wheels.sh
lucaseck, line 1: Not sure why this file is showing in the diff with no changes

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
crowecawcaw: I added this in here so that someone could make a dropdown menu for override options:
Screenshot 2024-05-22 at 7 30 55 AM

If default is a supported option, it can be added to every node and only changed for specific nodes that need overriding.

Someone that's adding a parameter as just an editable string probably wouldn't use this feature though.

scripts/build_wheels.sh
crowecawcaw: Ah, I made the file executable so we can run it.

test/deadline_submitter_for_houdini/unit/test_job_template.py
```diff
@@ -0,0 +1,184 @@
...
+    return mock_implementation
+
+
+@pytest.fixture(autouse=True)
+def init_mocks():
+    with patch(
+        "deadline.houdini_submitter.python.deadline_cloud_for_houdini.submitter._get_houdini_version",
+        Mock(return_value="19.5.435"),
+    ):
+        with patch(
```
joel-wong-aws: nit: for multiple patches, we can combine them into one `with` statement to reduce nesting:
```
with patch(
    "deadline.houdini_submitter.python.deadline_cloud_for_houdini.submitter._get_houdini_version",
    Mock(return_value="19.5.435")
), patch(
    "deadline.houdini_submitter.python.deadline_cloud_for_houdini.submitter._get_hip_file",
    Mock(return_value="/path/to/hip.hip"),
):
```

test/deadline_submitter_for_houdini/unit/test_job_template.py
```diff
@@ -0,0 +1,184 @@
...
+            "deadline.houdini_submitter.python.deadline_cloud_for_houdini.submitter._get_hip_file",
+            Mock(return_value="/path/to/hip.hip"),
+        ):
+            yield
+
+
+@patch("deadline.houdini_submitter.python.deadline_cloud_for_houdini.submitter._get_steps")
```
joel-wong-aws: nit: `_get_steps` is also mocked twice in this file, can it be moved into a fixture?

docs/user-guide.md
```diff
@@ -0,0 +1,30 @@
...
+
+By default, if a node is a geometry node with the "Initialize Simulation OPs" option enabled, it will render sequentially. Otherwise the node will render in parallel.
+
+### Overriding the render strategy
+
+You can override the render strategy be creating a `deadline_cloud_render_strategy` parameter with a value of either `SEQUENTIAL` or `PARALLEL`.
+
+To add a parameter to a node:
+1. Right click a node in the `out` context, select "Parameters and Channels" then click "Edit Paramter Interface".
+2. Under "Create Parameteres", select "Ordered Menu" and click the right arrow to add it to the "Existing Parameters" list.
```
joel-wong-aws: typo in `Create Parameters`

docs/user-guide.md
```diff
@@ -0,0 +1,30 @@
...
+    
+    b. Under the Menu tab, add menu items for:
+
+    | Token      | Label      |
+    | ---------- | ---------- |
+    | (empty)    | Default    |
+    | SEQUENTIAL | Sequential |
+    | PARALLEL   | Parallel   |
+4. Click Accept.
+5. Now in the parameter panel for your node, you'll find the new "Deadline Cloud Render Strategy" menu. Select an option to override the default submitter behavior.
```
joel-wong-aws: I believe what you're referring to is called the parameter editor, rather than parameter panel? https://www.sidefx.com/docs/houdini/network/parms.html

We should also apply whatever wording we go with to [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/docs/FAQ.md?plain=1#L8C4-L8C69) as well

crowecawcaw: I don't quite see the wording that needs to be aligned. Could you point it out?

docs/user-guide.md
crowecawcaw: Fixed this and a couple others.

docs/user-guide.md
joel-wong-aws: `5. Left click the Mantra node and navigate to "Rendering" > "Render"` -> `5. Left click the Mantra node and navigate to "Rendering" > "Render" in the parameter editor`

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -278,44 +315,64 @@ def _get_job_template(rop: hou.Node) -> dict[str, Any]:
...
-        task_data_contents = []
-        task_data_contents.append(f"render_node: {node['rop']}\n")
-        task_data_contents.append("frame: {{Task.Param.Frame}}\n")
-        task_data_contents.append(f"ignore_input_nodes: {ignore_input_nodes}\n")
-        # step
-        frame_range = "{start}-{stop}:{step}".format(**node)
+
+        task_data_dict = {"render_node": node["rop"], "ignore_input_nodes": ignore_input_nodes}
+
+        if node["render_strategy"] == RenderStrategy.SEQUENTIAL:
```
ddneilson: Would you mind refactoring this function to adhere to the sonarcloud guidance: https://sonarcloud.io/project/issues?resolved=false&pullRequest=149&id=aws-deadline_deadline-cloud-for-houdini&open=AY-WcNeZcKfnCYQ8pJrR

?

crowecawcaw: Refactored this function to reduce the branches. SonarCloud has some new complaints about existing code that I touched in the refactor. I don't think I'll fix those in this PR.

joel-wong-aws (dismissed):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
epmog: Is there a use-case for wanting to add it to all nodes? A user would have to script something to do this since they'd be editing every single node in the graph?

My understanding would be that we shouldn't provide a `default` option, because we will inevitably change what that default is for a particular node set-up and a user has no insight on what we think the default is.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
epmog: 1. In my above comment, I changed this to a `ValueError` instead of the generic `Exception`, can we do that here regardless of the above conversation?

2. nit: Maybe clean up the language/formatting a bit like:

```suggestion
            raise ValueError(
                f'The node "{node.path()}" has an unexpected value "{strategy_string}" for its "deadline_cloud_render_strategy" parameter. Ensure the value is {RenderStrategy.PARALLEL.value} or {RenderStrategy.SEQUENTIAL.value}'
            )
```

Not sure I want to add quotes around the enum values in-case a user accidentally adds quotes to the values.

joel-wong-aws (dismissed):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -66,10 +72,28 @@ def _get_steps(node: hou.Node):
...
+
+    if not separate_steps:
+        # render the node connected to the deadline cloud node
+        # and all its input nodes. The opposite of splitting it
+        # up each node by step
+
+        try:
+            connected_node = rop_steps[-1]
+        except IndexError:
+            return []
```
joel-wong-aws, line 88: nit: it would be better to raise an exception here (otherwise an exception will be raised down the chain), but based on the current code, this statement is actually unreachable, so not blocking

epmog (approved):
Thanks for this Stephen, looks great!

joel-wong-aws (approved):

Issue #116, opened 2024-04-12
Bug: Clicking "Parse Files" in Houdini clears manually added assets from the asset list


justinsaws:
### Expected Behaviour

The Houdini submitter shouldn't clear out manually added files from the asset lists.

### Current Behaviour

If you click "Parse Files", any manually added files are deleted from the asset list.

### Reproduction Steps

1. Manually add files to the asset list
2. Click "Parse Files"
3. Previous entries are removed.

### Code Snippet

https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L483-L492

epmog:
Other integrations that leverage the deadline-cloud UI track the automatically detected files compared to the manually entered files. As part of some of the other parse files fixes, once we're confident we're not adding problematic inputs, we can do the same here.

The code snippet here actually throws away everything when running the parsing, so we'll have to ensure we can identify the manually added ones vs. automatic since they're currently all just parms on the deadline-cloud rop node.

Can we add hidden metadata/invisible property of the parm that lets us know which ones are which?

PR #153, opened 2024-05-21
fix: parse files no longer removes manually added attachments

lucaseck:
#116 When a user clicked the `Parse Files` button it would remove any manually added attachment paths and only include what was detected in the scene.
When parsing the open scene for external files we now keep any difference between the detected paths and the paths currently stored in the file path parms. Any paths not included in the auto-detected set are displayed at the top of the lists as ones that have likely been manually added.

I did check if the font or something could be changed to distinguish more between auto-detected and manually added paths, but found no options for italicizing etc. In the future we may want separate list sections entirely.
Clicking parse files won't remove anything that was manually added
- Opened a scene and clicked parse files
- Added some manual paths
- Clicked parse files again and saw that the manually entered paths were moved to the top of the list and kept along with all of the auto-detected paths

New unit test cases were also added
No
No
----


src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
```diff
@@ -33,16 +33,27 @@ def _get_asset_references(rop_node: hou.Node) -> AssetReferences:
...
-        p = node.parm(ref)
-        while p.multiParmInstancesCount():
-            p.removeMultiParmInstance(0)
-        paths = sorted(list(getattr(asset_references, ref)))
-        p.set(len(paths))
-        for i, n in enumerate(p.multiParmInstances()):
-            n.set(paths[i])
+
+        auto_detected_paths = getattr(auto_detected_asset_references, ref)
+        manual_paths = getattr(assets, ref)
```
epmog: 1. Any reason we need to leverage getattr here?
2. Anything special happen if there's no assets?
3. What types are `assets`/`auto_detected_asset_references`? `hou.Parm`?

test/deadline_submitter_for_houdini/unit/test_assets.py
```diff
@@ -90,3 +92,79 @@ def test_get_recursive_output_directories(node_type: str, output_parm_name: str)
     node.parm.assert_called_once_with(output_parm_name)
     inner_node.parm.assert_called_with("vm_picture")
     assert out_dirs == {"/test/output/directory/mantra"}
+
+
+@pytest.mark.parametrize(
+    "scene_detected_assets, current_assets, expected_input_filenames, expected_input_directories, expected_output_directories",
+    [
+        (AssetReferences(), AssetReferences(), [], [], []),
```
epmog: can we leverage `pytest.param` here and provide `ids` to all of these test cases? Should make it easier to identify the intent of the test, and understand when things go wrong with the tests. Should also help idenitify gaps in the testing that we have for it. ie.

```
[ 
    pytest.param(AssetReferences(), AssetReferences(), [], [], [], id="no assets"),, 
    pytest.param(..., id="only auto detected inputs"),
    ...
]
```

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
```diff
@@ -33,16 +33,27 @@ def _get_asset_references(rop_node: hou.Node) -> AssetReferences:
...
+        manual_paths.difference_update(auto_detected_paths)
```
crowecawcaw: I think this logic will leave extra files in some cases. If someone has a large graph, clicks "Parse files", then removes some nodes and clicks "Parse files" again, won't the assets for the previously attached nodes remain even though they were automatically added and no longer needed? It sounded like the other submitters kept separate lists of auto-added and manually added attachments. Could we do the same?

epmog: Right, can we not leverage these two lists as hidden parms on the node and use that to populate the UI separately?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
lucaseck: 1. No special reason, it can be retrieved 3 times specifically instead of having to use getattr
2. I've tested before with no manual assets but I can reconfirm with no scene detected assets
3. Both are Houdini multiparms

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
lucaseck: You're correct I believe. I'll look into adding the hidden parms to avoid that case which isn't covered in the unit tests.

crowecawcaw (approved):

epmog (approved):

PR #154, opened 2024-05-21
doc: add FAQ, including how to fail immediately if textures are missing

joel-wong-aws:
A customer was confused about how to fail a render immediately if textures were missing (a cost/time-saving measure)

Added documentation about how to do so using Houdini's built-in options.

----

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -1,0 +1,9 @@
+## How can I fail Mantra renders immediately if textures are missing?
+
+There is a hidden feature in Mantra that can be used to fail the render immediately if textures are missing. To activate it:
+1. Right click the Mantra node and go to "Parameters and channels" > "Edit Parameter interface"
+2. In the "Create Parameters" section, select "Render Properties" and filter for "vm_abort_missing_texture"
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -0,0 +1,8 @@
+## How can I fail Mantra renders immediately if textures are missing?
+
+There is a hidden feature in Mantra that can be used to fail the render immediately if textures are missing. To activate it:
+1. Right click the Mantra node and go to "Parameters and channels" > "Edit Parameter interface"
+2. In the "Create Parameters" section, select "Render Properties" and filter for "vm_abort_missing_texture"
+3. Right click "Abort on missing texture (vm_abort_missing_texture)" and click "Install Parameter"
+4. Click "Accept"
+5. The parameter will appear as a checkbox on the Mantra node menu under "Rendering" > "Render" > "Abort on missing texture"
```
epmog: nit: the FAQ entry indicates it teaches you how to fail the render, but only tells you to install the parameter. We should probably have a last step of "checking" the checkbox to enable the feature, since just installing it does nothing.

epmog (approved):

crowecawcaw (approved):

PR #157, opened 2024-05-23
docs: add FAQ entry about incomplete asset definition errors

crowecawcaw:
Jobs submitted from this submitter produce errors while rendering on the farm even when rendering successfully. The errors are because Houdini on the worker does not have the submitter installed, but the submitted scene still contains the Deadline Cloud node which depends the submitter files.

The errors are safe to ignore since the Houdini node does not run as part of the render job.

Documents a point of potential confusion for customers.

n/a

Yes

No

----


epmog (dismissed):

docs/FAQ.md
```diff
@@ -7,3 +7,13 @@ There is a hidden feature in Mantra that can be used to fail the render immediat
...
+
+## Why do I get "incomplete asset definitions" errors while rendering?
+
+Jobs from this submitter that run in your farm may produce errors in the logs that look like:
+```
+The following node types are using incomplete asset definitions:
+  Driver/deadline_cloud
+```
+
+These errors are safe to ignore. This submitter exists as a node in your Houdini scene. When a worker in your farm loads the scene, the scene still contains the Deadline Cloud node, but the worker may not have the submitter installed. Because the worker does not have the files needed to run the Deadline Cloud node, it logs "incomplete asset definition" errors. The Deadline Cloud node itself is not rendered as part of the job though, so these errors can be ignored.
```
joel-wong-aws: `This submitter` isn't clear what it is referring to, how about `The AWS Deadline Cloud submitter` instead?

epmog (approved):

joel-wong-aws (approved):

PR #158, opened 2024-05-27
fix: add missing submission telemetry event

lucaseck:
Missing a telemetry event for submissions
Add the event
No more missing telemetry
Ran a submission and saw the event record
No
No
----


marofke (dismissed):

epmog:
Nice! Should we also be adding one for exceptions raised in submission? ie. lines 674-678

Would look something like: https://github.com/aws-deadline/deadline-cloud/blob/041550c40002e3a2d1b77a77530318212cf10b5d/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py#L487-L492

marofke (approved):

epmog (approved):

PR #162, opened 2024-05-31
feat: capture files with time-based variables for job attachments

### LINKED ISSUE #125: Bug: filepaths with $F token in the input are not expanded to include all files for a frame range
### Expected Behaviour

If a scene contains an external filepath reference using a [frame expression](https://www.sidefx.com/docs/houdini/render/expressions.html) such as $F4 then all files needed for the frame range being submitted should be included in the upload.

### Current Behaviour

If a filepath is specified with $F4 specified then the filepath is only evaluated and added to the uploads for the currently selected frame.

### Reproduction Steps

To reproduce you can:
- create a new scene with some geometry(a cube)
- create some number of sample png files named like `eyes_0001.png` under the $HIP directory, one for each frame you're attempting to render
- create a new material
- set the material Texture Map path to something like `$HIP/eyes_$F4.png`
- add the material to your created geometry
- create new mantra and deadline cloud nodes
- click parse files on the deadline node
- you should see an input file item like `$HIP/eyes_$F4.png`
- ensure a frame range is selected, not only the current frame to render
- submit the job

In the generated asset manifest file there will only be a single input filename for the currently selected frame such as `C:/Users/<USER>/Desktop/HoudiniFrameTesting/eyes_0001.png` not a file for each frame in the frame range.

### Code Snippet

https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py#L461 can likely be modified to expand any filenames that include tokens like $F for the entire frame range that is set to render.

epmog:
Resolves #125 

The deadliine-cloud rop node was evaluating parms at the current frame. So any filepaths that contained time-based variables would only have 1 of those files captured rather than the whole sequence

There were a few solutions available, each with various downsides.

1. Evaluate the parm with evalAsStringAtFrame`, would require us to figure out all the relevant frame numbers and adjust for offsets.
    * this requires us to get it perfectly correct - which is definitely higher effort

2. Grab the directory before the frame variable
    * anecdotal evidence seems to indicate frames are mostly in the file name, but users are not restricted to having them in directories. So automatically uploading whole directories without insight as to why they're added leads to the potential of a lot of extraneous files uploaded. 

3. Replace the relevant variables with globs and and let the stdlib take care of finding them.

I went with replacing the time-based variables for input filenames with globs to allow us to do some naive pattern matching. The code itself is then pretty simple with avenues to improve it later.
* Downside here is there may be inadvertent files that get included because they match the glob pattern and not the time variable. Future improvement could be to filter out files that matched the globs but not the pattern
* Input directories are being ignored in this change, since the solution can be equivalent to going up a directory, which can be pretty dangerous while multiple levels of directories down from frames seem like a pretty rare case.

We get a LOT more of the files users want to be included with the job!

Added a bunch of unit tests to capture the pattern replacement and globbing functionality. 

```
hatch run fmt
hatch build
hatch run lint
hatch run test
```

Test coverage is now at ~67%
```
Required test coverage of 65.0% reached. Total coverage: 67.59%
```

I'll need to retest the node within Houdini again after some rebasing.

N/A

Fixing!

----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
@@ -1,5 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+import glob
 import os
+import re
 
 from deadline.client.job_bundle.submission import AssetReferences
 
@@ -22,6 +24,50 @@ def _get_hip_file() -> str:
     return hou.hipFile.path()
 
 
+def _houdini_time_vars_to_glob(path: str) -> str:
+    """
+    Given a path, replace any houdini time-based variables or expressions with globs
+    to capture complete file sequences for job attachments. If there is no time-based
+    parameterization, then the original string is returned.
+
+    These are the useful timebased variables that can be included in files.
+    ref: https://www.sidefx.com/docs/houdini/render/expressions.html
+    """
+    time_based_variable_patterns: tuple = (
+        "(`.*`)",  # an expression, like python!
+        "(\$FF)",  # current fractional frame number
+        "(\${FF})",
+        "(\$F\d*)",  # current frame number, only variable with digits that follow
+        "(\${F\d*})",
+        "(\$T)",  # current time
+        "(\${T})",
+        "(\$SF)",  # current simulation timestep number
+        "(\${SF})",
+        "(\$ST)",  # current simulation time
+        "(\${ST})",
+    )
+    time_var_pattern = "|".join(time_based_variable_patterns)
+
+    return re.sub(time_var_pattern, "*", path)
+
+
+def _get_evaluated_glob_path(parm, globbed_path: str) -> str:
+    """Given a parm that contains a path, temporarily set it to a globbed path
+    and perform an evaluation to remove all other variables, return the value
+    and ensure that we put the original value back.
+
+    Rationale: Houdini does not appear to provide a way to evaluate an arbitary
+    string, only as a Parm. Thus, we can use this function to put a value into
+    a parm to then evaluate.
+    """
+    orig_path = parm.unexpandedString()
+    try:
+        parm.set(globbed_path)
+        return parm.eval()
+    finally:
+        parm.set(orig_path)
+
+
 def _get_asset_references(rop_node: hou.Node) -> AssetReferences:
     """
     Get the current paths stored in the parms backing the UI and return them as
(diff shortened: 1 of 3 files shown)


src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
```diff
@@ -22,16 +24,54 @@ def _get_hip_file() -> str:
     return hou.hipFile.path()
 
 
+def _houdini_time_vars_to_glob(str_: str) -> str:
```
lucaseck: nit: This is expecting explicitly a filename as the input `str_ -> filename or path`?

test/deadline_submitter_for_houdini/unit/test_assets.py
```diff
@@ -212,3 +219,248 @@ def test_parse_files_manually_added(
...
+            "sequence.0.png",
+            "sequence.1.png",
+        ],
+        id="Python expression with a frame variable within",
+    ),
+]
+
+
+@pytest.mark.parametrize(("pattern", "filenames"), filenames_with_frames)
+def test_filenames_with_frames(tmpdir, pattern: str, filenames: list[str]):
```
lucaseck, line 386: Might be missing it, but it looks like none of these test cases include additional files that **shouldn't** be included in the matching. Can an additional parameter be added for `expected_files `and additional files added to the `filenames` that shouldn't/won't be matched?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
```diff
@@ -22,16 +24,54 @@ def _get_hip_file() -> str:
...
+            continue
+
+        # This is a work-around to avoid evaluating the offsets for
+        # scene introspection. It's possible that this will capture
+        # extra files since a glob will not enforce numbers, but
+        # handles well-defined file path patterns
+        for file in glob.iglob(filepath_glob):
+            asset_references.input_filenames.add(file)
+
     asset_references.input_directories.update(
```
lucaseck, line 96: We don't expand tokens in directory names?

epmog: This is addressed by a comment in the PR description, but we can chat about it here if you'd like

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
epmog: It'll work on any string, but the inputs we're dealing with are assumed to be path-like

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
lucaseck: Ah missed the comment. So there is concern about how to handle directory traversal that may happen as part of the expansion? What is the expectation then for users in terms of tokens in directories then, is that a complete non-starter and it shouldn't be done at all?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
lucaseck: Nit was basically the parameter name isn't clear about the expectation that the inputs here are file paths like the comment string says.

crowecawcaw (dismissed):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
epmog: My current take on it, is that it's more rare than being in filenames - though I'm sure it'll come up eventually. To avoid an absolute explosion of files being added (ie. if the last folder is just the token) we actually have to pattern match the results to the token after globbing.

I'll put a thing in the FAQ about this

lucaseck: That sounds good to me. If it becomes an issue for people and is reported we can revisit.

crowecawcaw (approved):

lucaseck (approved):

PR #190, opened 2024-11-01
fix: detect ROP USD Render output directories

### LINKED ISSUE #191: Bug: USD Render ROP Output Directories Are Not Detected
### Describe the bug

When using a USD Render ROP in a scene the output directories are not detected even if the override is set.

### Expected Behaviour

The Deadline Cloud node should detect output directories needed by the USD Render ROPs

### Current Behaviour

Output directories are not being detected and added to the job attachments list

### Reproduction Steps

1. Create a new scene
2. Add a USD Render ROP in /out
3. Set the Override Output Image to a path to some output file
4. Connect 2. to a Deadline Cloud node in /out
5. In the Deadline Cloud job attachments tab click on parse files, the output directory list will not be updated

### Environment

At minimum:
1. Operating system: Windows 10
2. Version of Houdini: 19.5/20.0/20.5
3: Version of this package: 0.7.3 
4. N/A

lucaseck:
Fixes: #191 

* Output directories were not being detected from USD Render ROPs
* The output override field was not being taken into account when detecting USD Render output directories
* Output directories were not being detected from specified LOP Paths in the USD Render nodes

* Add a new handler for USD Render nodes(both LOP and ROP USD Render nodes)
* Add logic for taking the override output path if one is provided in the nodes
* Add logic for taking the specified LOP Path node as the node to detect output directories from if one is provided
* Fixed an additional bug where we weren't checking that a USD Prim was invalid(doesn't exist) or not

More output directory detection cases are handled

Manually tested by creating a scene with both a ROP USD Render node and a LOP USD Render node. Verified all of the cases for overrides, LOP Paths, regular inputs to the nodes worked for detecting the output directories.

Unit tests passed

No

No

----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
@@ -265,17 +265,43 @@ def _husk_outputs(node: hou.Node) -> set[str]:
     of the USD stage input
     """
     output_directories: set[str] = set()
+
     for n in node.inputs():
         try:
             products = n.stage().GetPrimAtPath("/Render/Products")
+            # GetPrimAtPath returns an invalid null Prim object
+            # if a Prim doesn't exist at the specified path
+            if not products.IsValid():
+                continue
         except Exception:
             # no products found in standard namespace
-            return output_directories
+            continue
+
         for child in products.GetChildren():
             if child.GetTypeName() == "RenderProduct":
                 product_name_attr = child.GetAttribute("productName")
                 path = product_name_attr.Get(0)
                 output_directories.add(os.path.dirname(path))
+    return output_directories
+
+
+def _usd_render_outputs(node: hou.Node) -> set[str]:
+    """Obtain list of outputs by checking in priority order for:
+    1) The override output image value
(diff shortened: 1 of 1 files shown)


epmog (approved):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
```diff
@@ -279,6 +285,26 @@ def _husk_outputs(node: hou.Node) -> set[str]:
     return output_directories
 
 
+def _usd_render_outputs(node: hou.Node) -> set[str]:
```
miabatta, line 288: Is there benefit in adding a regression test for this? Depending on how the Houdini mock works it might not add anything of use, so I'm just curious what you think

lucaseck: A unit test could be added. Personally I don't know how valuable it would be given how much has to be mocked and crafted to see the desired behaviour which at the end of the day would mostly just be testing does the if clause work. I would prefer this gets covered in the near future by integration tests that are actually running within Houdini and able to test these workflows e.g. USD Renders

miabatta (approved):

PR #192, opened 2024-11-06
feat: installer component support for Houdini 20.0 and 20.5

### LINKED ISSUE #123: Bug: Houdini Submitter User Install does not work with certain usernames
### Expected Behaviour

DeadlineCloudSubmitter for Houdini should work as a User Install.

### Current Behaviour

User install doesn't work and Houdini can't find site-packages for DeadlineCloud

### Reproduction Steps

Be a user who's username starts with a 'b'. Install Houdini and the DeadlineCloudSubmitter as a user install (rather than system wide).

Launch Houdini, notice the submitter is not loaded. Check the `sys.path` inside Houdini's python interpreter:

```
>>> sys.path
['C:/PROGRA~1/SIDEEF~1/HOUDIN~1.805/python39/lib/site-packages-ui-forced', 'C:/PROGRA~1/SIDEEF~1/HOUDIN~1.805/python39/lib/site-packages-forced', '', 'C:\\Users\x08lantrev\\DeadlineCloudSubmitter\\Submitters\\Houdini\\python', ......
```

I suspect some weird path escape/replace is happening hence the `C:\\Users\x08lantrev`. And this is what is causing the issue. 

Re-installing with a system wide install and the full path the the submitter packages location is `C:\Program Files\DeadlineCloudSubmitter\Submitters\Houdini\python` and the Deadline node is available in Houdini. 

### Code Snippet

n/a

lucaseck:
Fixes: #123 

The submitter installer component was written in a way that it would only work for Houdini 19.5 installations.
- Generalized portions of the installer component so that it will install for 19.5, 20.0 and 20.5 while also hopefully making it easier to update in the future for new versions
  - Each version of Houdini is now a subcomponent and the overall Houdini component is a component group
  - There did not appear to be a way to have the folder section copy to multiple destinations so instead the package file is copied to a temp location before then being copied over to the Houdini packages folders for the selected versions. The swapping of the INSTALL_DIR_PLACEHOLDER has now been moved to the folder actionList which runs immediately after the file is copied to the temp location before it's then copied to the Houdini package folders.
- Changed the substitution of `INSTALL_DIR_PLACEHOLDER` to use unix paths on Windows to fix #123 so that there are no weird escape issues with certain characters and unescaped backslashes on Windows.

Selection of the version of Houdini will look like below:
image

New CLI flags to select the components:
image

- A future release of the installer will use the new component and be able to install for Houdini 20.0 and 20.5

- Built dev builds of the installer for Windows and Linux
- Ran through the installer on both OSes in GUI mode and verified that I could select individual Houdini versions to install for and validated that nothing was being changed for the other versions. 
- Ran through the installer on both OSes in CLImode and verified that I could select individual Houdini versions via flags during installation and validated that nothing was being changed for the other versions. 
- I verified that if the Houdini package directory hadn't been made yet the installer would create it(the packages directory doesn't exist by default and the user pref directory won't until Houdini has been launched at least once).
- Verified that on unisntallation the package files were removed
- Verified that on installation over an existing installation the package files were still updated
- Verified on each OS that I could use the submitter in Houdini 19.5, 20.0 and 20.5

No

No

----

### DIFF
--- a/install_builder/deadline-cloud-for-houdini.xml
+++ b/install_builder/deadline-cloud-for-houdini.xml
@@ -1,7 +1,7 @@
-<component>
+<componentGroup>
     <name>deadline_cloud_for_houdini</name>
-    <description>Deadline Cloud for Houdini 19.5</description>
-    <detailedDescription>Houdini plugin for submitting jobs to AWS Deadline Cloud. Compatible with Houdini 19.5.</detailedDescription>
+    <description>Deadline Cloud for Houdini</description>
+    <detailedDescription>Houdini plugin for submitting jobs to AWS Deadline Cloud. Compatible with Houdini 19.5, 20.0 and 20.5</detailedDescription>
     <canBeEdited>1</canBeEdited>
     <selected>0</selected>
     <show>1</show>
@@ -19,7 +19,7 @@
         </folder>
         <folder>
             <description>Package</description>
-            <destination>${houdini_user_pref_dir_default}/packages</destination>
+            <destination>${installdir}/tmp/packages</destination>
             <name>houdinipackage</name>
             <platforms>all</platforms>
             <distributionFileList>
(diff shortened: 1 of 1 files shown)


erico-aws (approved):

miabatta (approved):

miabatta:
Needs rebase then should be good 👍

PR #203, opened 2024-12-09
fix(installer): fix installer system installs

### LINKED ISSUE #197: Bug: Linux: Submitter Does Not Work When Doing System Install
### Describe the bug

On Linux when running the submitter installer, if installed for all users you need to run the installer as root. If you do this, the installer will always put the houdini package files in the installing user's home directory. Because of this:

1. When a normal user opens houdini, there will be no package file on their houdini path and the plugin will not be loaded
2. They cannot even put it on their path because they wouldn't have access to root's home directory

### Expected Behaviour

The package files should be installed somewhere that all users have access to

### Current Behaviour

See above

### Reproduction Steps

Using the submitter installer, install the houdini submitter on linux for all users. Open Houdini and observe that the plugin has not loaded.

### Environment

At minimum:
1. Operating system: RHEL 9
2. Version of Houdini: All
3: Version of this package: 0.7.4

lucaseck:
Fixes: #197 

The installer was not installing deadline cloud for Houdini correctly when doing a system wide install. The package file was only ever being placed in the default Houdini user pref directory.

Swapped the installer component to use the env var [`HOUDINI_PACKAGE_DIR`](https://www.sidefx.com/docs/houdini/ref/plugins.html#using_packages) so that both user and system installs work. A nice additional benefit is that we don't have to have multiple subcomponents to select for each version of Houdini anymore.

Future installers will work to install Deadline Cloud for Houdini with both user and system installations.

Manually built both Windows and Linux installers and:
* Ran a user install and verified HOUDINI_PACKAGE_DIR was being set and the submitter node loaded and could submit in Houdini 19.5 and 20.5
* Ran a system install and verified HOUDINI_PACKAGE_DIR was being set and the submitter node loaded and could submit in Houdini 19.5 and 20.5

On just Windows I verified that I could install over an existing installation with the current method of putting the package file in the default user pref directory. The previous installation package file was deleted from the default user pref directory. The new installation then set HOUDINI_PACKAGE_DIR. Verified in Houdini's Asset Manager that the correct path to the Node definition was being picked up.

Yes

No

No

----

### DIFF
--- a/install_builder/deadline-cloud-for-houdini.xml
+++ b/install_builder/deadline-cloud-for-houdini.xml
@@ -1,7 +1,7 @@
 <componentGroup>
     <name>deadline_cloud_for_houdini</name>
-    <description>Deadline Cloud for Houdini</description>
-    <detailedDescription>Houdini plugin for submitting jobs to AWS Deadline Cloud. Compatible with Houdini 19.5, 20.0 and 20.5</detailedDescription>
+    <description>Deadline Cloud for Houdini 19.5-20.5</description>
+    <detailedDescription>Houdini plugin for submitting jobs to AWS Deadline Cloud. Compatible with Houdini 19.5-20.5</detailedDescription>
     <canBeEdited>1</canBeEdited>
     <selected>0</selected>
     <show>1</show>
@@ -19,7 +19,7 @@
         </folder>
         <folder>
             <description>Package</description>
-            <destination>${installdir}/tmp/packages</destination>
+            <destination>${houdini_packagedir}/</destination>
             <name>houdinipackage</name>
             <platforms>all</platforms>
             <distributionFileList>
@@ -30,13 +30,20 @@
             <actionList>
                 <setInstallerVariable name="deadline_package_file_name" value="deadline_submitter_for_houdini.json" />
                 <substitute>
-                    <files>${installdir}/tmp/packages/${deadline_package_file_name}</files>
+                    <files>${houdini_packagedir}/${deadline_package_file_name}</files>
                     <type>exact</type>
                     <encoding>utf-8</encoding>
                     <substitutionList>
                         <substitution pattern="INSTALL_DIR_PLACEHOLDER" value="${installdir.unix}" />
                     </substitutionList>
                 </substitute>
+                <fnAddPathEnvironmentVariable>
+                    <progressText>Setting HOUDINI_PACKAGE_DIR</progressText>
+                    <name>HOUDINI_PACKAGE_DIR</name>
+                    <value>${houdini_packagedir}</value>
+                    <scope>${installscope}</scope>
+                    <insertAt>end</insertAt>
+                </fnAddPathEnvironmentVariable>
             </actionList>
         </folder>
         <folder>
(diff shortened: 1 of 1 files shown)


install_builder/deadline-cloud-for-houdini.xml
```diff
@@ -51,57 +58,12 @@
...
-                <fnCopyHoudiniPackageFile houdiniVersion="20.5" houdiniPackageDir="${houdini_user_pref_dir_default}" />
-            </postInstallationActionList>
-        </component>
-    </componentList>
     <initializationActionList>
         <setInstallerVariable name="all_components" value="${all_components} deadline_cloud_for_houdini" />
     </initializationActionList>
     <readyToInstallActionList>
         <setInstallerVariable name="houdini_installdir" value="${installdir}/Submitters/Houdini" />
+        <setInstallerVariable name="houdini_packagedir" value="${houdini_installdir}/packages" />
```
edwards-aws: Do we need to remove `packages` here? The for `HOUDINI_PACKAGE_DIR` envvar the Houdini docs say

> Houdini will process the package files directly in the directory specified in HOUDINI_PACKAGE_DIR, do not add a packages directory under these folders.

edwards-aws (changes requested):
Thanks Lucas! Just one quick question.

install_builder/deadline-cloud-for-houdini.xml
```diff
@@ -30,13 +30,20 @@
...
+                    <files>${houdini_packagedir}/${deadline_package_file_name}</files>
                     <type>exact</type>
                     <encoding>utf-8</encoding>
                     <substitutionList>
                         <substitution pattern="INSTALL_DIR_PLACEHOLDER" value="${installdir.unix}" />
                     </substitutionList>
                 </substitute>
+                <fnAddPathEnvironmentVariable>
+                    <progressText>Setting HOUDINI_PACKAGE_DIR</progressText>
+                    <name>HOUDINI_PACKAGE_DIR</name>
```
edwards-aws, line 42: It seems like this is a Houdini wide setting, is it possible that this can conflict with other plugins or settings the customer might have already set for themselves?

edwards-aws (changes requested):
Sorry one more question

install_builder/deadline-cloud-for-houdini.xml
lucaseck: We are just appending to the end of the variable if it exists, so there should be no impact on existing settings.

install_builder/deadline-cloud-for-houdini.xml
lucaseck: I think what they're meaning by that is that you don't need to add a `packages` subdirectory under the HOUDINI_PACKAGE_DIR like there is for the default search paths. They'll just load whatever is at the top level pointed to by HOUDINI_PACKAGE_DIR. I've chosen to name it packages just to make it match the existing setup and explicitly point the env var to that level.

install_builder/deadline-cloud-for-houdini.xml
edwards-aws: Ahhh, it's a Path type variable. Makes sense, thanks!

install_builder/deadline-cloud-for-houdini.xml
edwards-aws: We might want to choose something different. In the Houdini docs they italicize _do not_ so it seems like they're really emphasizing that this is something we shouldn't do. I don't think they'd do that without good reason.

It's unclear to me from the docs why we wouldn't want to use `packages`. It's also unclear if we shouldn't have something in that variable just called `packages` or if the final directory in a path shouldn't be `packages`.

I think we should err on the side of caution here and chose something different unless we have a really good reason not to.

lucaseck: I think it's just a warning so that people aren't confused because the default search locations all expect to be under packages and you don't need a packages folder under $HOUDINI_PACKAGE_DIR like there is for HOUDINI_USER_PREF_DIR and HSITE.
```
$HOUDINI_USER_PREF_DIR/packages

$HSITE/houdini‹major›.‹minor›/packages (for example, $HSITE/houdini18.0/packages)
```

No harm in changing it though either way so I'll put up a quick change

lucaseck: Put up a small change to turn it into `deadline_package` as the folder name holding the package file.

edwards-aws (approved):
looks good!

epmog (approved):

PR #219, opened 2025-02-05
chore: update README

joel-wong-aws:
The instructions about how to setup the Deadline Cloud submitter for Houdini and submit jobs was unclear. Added more detailed instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,6 @@ There are two installation options:
 
 For most setups, you will also want to install the [Deadline Cloud monitor][deadline-cloud-monitor].
 
-[Houdini Render nodes (ROP)]: https://www.sidefx.com/docs/houdini/nodes/out/index.html
-[deadline-cloud-monitor-setup]: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html#install-deadline-cloud-monitor
-[aws-cli-credentials]: https://docs.aws.amazon.com/cli/v1/userguide/cli-chap-authentication.html
-
 ### Using the Submitter
 
 To use the Houdini submitter:
@@ -65,7 +61,7 @@ To use the Houdini submitter:
 7. Enter any job settings in the node editor, usually in the upper right side of Houdini.
 8. In the bottom right of the node editor, choose "Submit".
 
-The Deadline Cloud submission will automatically parse the connected `/out` network tree and submit each node as a step in the job maintaining the dependency tree.
+The Deadline Cloud submission will automatically parse the connected `/out` network tree and submit each node as a step in the job maintaining the dependency tree. Using non-default render networks other than `/out` is also supported.
 
 ### Adaptor
 
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 [![python](https://img.shields.io/pypi/pyversions/deadline-cloud-for-houdini.svg?style=flat)](https://pypi.python.org/pypi/deadline-cloud-for-houdini)
 [![license](https://img.shields.io/pypi/l/deadline-cloud-for-houdini.svg?style=flat)](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/LICENSE)
 
-AWS Deadline Cloud for Houdini is a Python package that supports creating and running SideFX Houdini jobs within [AWS Deadline Cloud](deadline-cloud).
+AWS Deadline Cloud for Houdini is a Python package that supports creating and running SideFX Houdini jobs within [AWS Deadline Cloud][deadline-cloud].
 It provides both the implementation of a Houdini plug-in for your workstation that helps you offload the computation for your rendering workloads
-to [AWS Deadline Cloud](deadline-cloud) to free up your workstation's compute for other tasks, and the implementation of a command-line
+to [AWS Deadline Cloud][deadline-cloud] to free up your workstation's compute for other tasks, and the implementation of a command-line
 adaptor application based on the [Open Job Description (OpenJD) Adaptor Runtime][openjd-adaptor-runtime] that improves AWS Deadline Cloud's
 ability to run Houdini efficiently on your render farm.
 
@@ -17,6 +17,7 @@ ability to run Houdini efficiently on your render farm.
 [openjd-adaptor-runtime-lifecycle]: https://github.com/OpenJobDescription/openjd-adaptor-runtime-for-python/blob/release/README.md#adaptor-lifecycle
 [service-managed-fleets]: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/smf-manage.html
 [default-queue-environment]: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/create-queue-environment.html#conda-queue-environment
+[deadline-cloud-submitter]: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html
 [deadline-cloud-monitor]: https://docs.aws.amazon.com/deadline-cloud/latest/userguide/monitor-onboarding.html
 
 ## Compatibility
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -47,17 +38,44 @@ Before submitting any large, complex, or otherwise compute-heavy Houdini render
...
-### Houdini Adaptor
+### Install the submitter using the official Deadline Cloud submitter installer
+
+The [official Deadline Cloud submitter installer][deadline-cloud-submitter] for Windows and Linux include the Houdini submitter. After installing, you can use the submitter in the Houdini.
+
+### Manually installing the submitter
+
+Manual installation is supported on Windows, Mac, and Linux.
+
+To install the submitter plug-in, refer to [DEVELOPMENT.md](DEVELOPMENT.md) section `Submitter Development Workflow`. Also refer to Deadline Cloud's [documentation](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html) on how to launch and use the Houdini submitter after installation.
```
erico-aws: idk if you need this last sentence. Seems like you've got the full instructions right below here. Maybe change it to "for more details refer to the Deadline Cloud documentation."

README.md
```diff
@@ -17,25 +17,16 @@ ability to run Houdini efficiently on your render farm.
...
 ## Compatibility
 
 This library requires:
 
-1. Houdini 19.5, 20.0 or 20.5
+1. Houdini 19.5, 20.0, or 20.5
 1. Python 3.9 or higher; and
 1. Linux, Windows, or a macOS operating system.
 
-## Versioning
```
lucaseck, line 29: The other repositories([Maya](https://github.com/aws-deadline/deadline-cloud-for-maya?tab=readme-ov-file#versioning),[Nuke](https://github.com/aws-deadline/deadline-cloud-for-nuke?tab=readme-ov-file#versioning),[Blender](https://github.com/aws-deadline/deadline-cloud-for-blender?tab=readme-ov-file#versioning)) all seem to align on having the versioning information front and center in the README. Personally I like keeping it in the README since it's the most likely place people will find it and it's not specific to just development with the package. Users who only consume the package will want to know that we make breaking changes between minor releases and may not care about looking in the DEVELOPMENT.md

README.md
```diff
@@ -47,17 +38,44 @@ Before submitting any large, complex, or otherwise compute-heavy Houdini render
...
+This package provides a Houdini render output node (ROP) that creates jobs for AWS Deadline Cloud using the [AWS Deadline Cloud client library][deadline-cloud-client]. Based on the loaded scene it determines the files required, allows the user to specify render options, and builds an [OpenJD template][openjd] that defines the workflow.
 
-To install the submitter plug-in, refer to [DEVELOPMENT.md](DEVELOPMENT.md) section `Submitter Development Workflow`. Also refer to Deadline Cloud's [documentation](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html) on how to launch and use the Houdini submitter after installation.
+There are two installation options:
+1. Windows or Linux: Install the submitter using the official Deadline Cloud submitter installer for Windows or Linux
+2. Windows, Mac, or Linux: Manually install the submitter
+
+For most setups, you will also need to install the [Deadline Cloud monitor][deadline-cloud-monitor].
```
lucaseck: super nit: `need` vs `want` AFAIK users never 100% need the monitor to be installed, but it's strongly recommended for credentials etc. Just don't want to give the wrong impression of what's strictly required

README.md
```diff
@@ -47,17 +38,44 @@ Before submitting any large, complex, or otherwise compute-heavy Houdini render
...
+
+### Manually installing the submitter
+
+Manual installation is supported on Windows, Mac, and Linux.
+
+To install the submitter plug-in, refer to [DEVELOPMENT.md](DEVELOPMENT.md) section `Submitter Development Workflow`. Also refer to Deadline Cloud's [documentation](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html) on how to launch and use the Houdini submitter after installation.
+
+### Using the Submitter
+
+To use the Houdini submitter:
```
lucaseck, line 54: nit: I'm wondering if we should include a note as well that the Deadline Cloud Node isn't limited to only being used in /out. Haven't tested in a while, but I believe you can put a ROP subnetwork somewhere other than /out and still use the submitter node just fine. e.g. put a ROP subnetwork in /geo and place the Deadline Cloud Node in there

README.md
joel-wong-aws: Updated to `want`!

README.md
joel-wong-aws: Thanks, removed!

README.md
joel-wong-aws: I think we should focus on the officially supported/tested flows here in the README. If users want to explore different approaches, that's fine, but this is the canonical way to do it.

README.md
joel-wong-aws: I don't think versioning is critical information for most readers of this README. If a user is onboarding for the first time, they likely don't want or need to know about our versioning structure; they likely just want to run through the instructions for setup.

I understand your point that versioning is broader than just for development, but if users are thinking about upgrading (where versioning is most relevant), I expect the more likely place they will look for and find this information is directly in the releases, e.g. [0.7.0](https://github.com/aws-deadline/deadline-cloud-for-houdini/releases/tag/0.7.0). With that approach, they also don't have to remember our versioning structure, they just have to look for the clear BREAKING CHANGES header.

I still think this should go into DEVELOPMENT.md because the only time there is something actionable to do with the information is during development, but I think at a minimum this section should go to the bottom of the README.

erico-aws (dismissed):

README.md
lucaseck: I'd prefer to keep it in the README and keep the alignment with the other repositories unless all others all also moving it. Best to keep the consistency IMO.

README.md
```diff
@@ -47,17 +38,45 @@ Before submitting any large, complex, or otherwise compute-heavy Houdini render
...
 
-This package provides a Houdini ROP node that creates jobs for AWS Deadline Cloud using the [AWS Deadline Cloud client library][deadline-cloud-client]. Based on the loaded scene it determines the files required, allows the user to specify render options, and builds an [OpenJD template][openjd] that defines the workflow.
+This package provides a Houdini render output node (ROP) that creates jobs for AWS Deadline Cloud using the [AWS Deadline Cloud client library][deadline-cloud-client]. Based on the loaded scene it determines the files required, allows the user to specify render options, and builds an [OpenJD template][openjd] that defines the workflow.
 
-To install the submitter plug-in, refer to [DEVELOPMENT.md](DEVELOPMENT.md) section `Submitter Development Workflow`. Also refer to Deadline Cloud's [documentation](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html) on how to launch and use the Houdini submitter after installation.
+There are two installation options:
+1. Windows or Linux: Install the submitter using the official Deadline Cloud submitter installer for Windows or Linux
+2. Windows, Mac, or Linux: Manually install the submitter
```
epmog: We should probably just link to the service documentation here for number 1, and link to the development "how to install" for number 2. They don't need the separate sections down below that just say the exact same info.

README.md
epmog: It's a ROP node - it works in render networks. We mostly use the default render network (`/out/`), but Houdini gives a lot of freedom in how you design your pipelines/trees.

https://www.sidefx.com/docs/houdini/nodes/out/index.html

> Houdini creates a default render network at /out. Items in the top-level Render menu let you create and edit render nodes in this network. You can also create your own ROP networks at any level of the scene. For example, you might embed a render node in an asset as part of an ability to generate its own textures.

In which case, something like this would work:

> In the Network Editor, usually in the lower right side of Houdini, navigate to a render network (example `/out`)

epmog (changes requested):

README.md
joel-wong-aws: Updated!

README.md
joel-wong-aws: I have moved this to the bottom of the README instead!

epmog (approved):

lucaseck (approved):

PR #226, opened 2025-02-17
chore: fixed dev install to support x86_64 or arm64

### LINKED ISSUE #218: Bug: Wrong architecture for newer versions of Houdini on Apple Silicon
### Describe the bug

Apple silicon uses arm64 but the hatch install process hardcodes x86 binaries which crashes the Houdini Submitter

To fix the issue. Delete these two lines: https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/scripts/install_dev_submitter.py#L124-L125

### Expected Behaviour

The hatch scripts should install the correct platform for the computer it's on.

### Current Behaviour

x86 binaries are hardcoded: https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/scripts/_project.py#L68-L69

### Reproduction Steps

run the submitter install steps in the readme on an Apple Silicon computer.

### Environment

At minimum:
1. Operating system macOS Apple Silicon
2. Version of Houdini: 20.5
3. Version of this package 0.7.5

epmog:
Fixes: #218

We were only ever grabbing the x86_64 native dependencies for the dev install. This causes issues for users running any other CPU architecture. In the case of macOS, this specifically caused issues with Apple silicon if Houdini was not running under Rosetta.

Added a `--cpu-arch` option that defaults to your current CPU architecture. A user that is emulating Houdini with a different architecture will have to specify this value to match. 

While I was here, I also added python 3.13 to the CIs and as a supported version, even though there isn't a Houdini version that uses python 3.13 yet.

Users running Houdini made for Apple silicon can now use `houdini run install` and have it work out of the box.

Ran the hatch run install on macOS specifying the different options and no option.

Nope

----

### DIFF
--- a/scripts/_project.py
+++ b/scripts/_project.py
@@ -5,12 +5,18 @@ from __future__ import annotations
 import subprocess
 import sys
 
+from enum import Enum
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any, Optional
 
 
 ADAPTOR_ONLY_DEPENDENCIES = {"openjd-adaptor-runtime"}
+
+
+class CPUArch(Enum):
+    X86_64: str = "x86_64"
+    ARM64: str = "arm64"
 
 
 def get_project_dict(project_path: Optional[Path] = None) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


scripts/_project.py
```diff
@@ -62,12 +68,23 @@ def get_git_root() -> Path:
...
 
 
-def get_pip_platform(system_platform: str) -> str:
+def get_pip_platform(system_platform: str, cpu_arch: CPUArch = CPUArch.X86_64) -> str:
     if system_platform == "Windows":
-        return "win_amd64"
-    elif system_platform == "Darwin":
-        return "macosx_10_9_x86_64"
-    elif system_platform == "Linux":
-        return "manylinux2014_x86_64"
```
epmog, line 71: This is an alias for the new value `manylinux_2_17_x86_64`: https://peps.python.org/pep-0600/

leongdl: Seems like all of our repos that support OSX should be arm vs x86 aware? eg: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/c4f7d3a42cdc7b5166a5e6e6ba08eca54154e076/scripts/create_adaptor_packaging_artifact.sh#L31

Should we log tickets for each team so they do it themselves?

joel-wong-aws (approved):

lucaseck (approved):

scripts/_project.py
epmog: That's specifically for adaptors and is only a potential issue on windows since it has pywin32 as a native dependency.

Eventually that may become something that teams need to fix, but it definitely doesn't need doing right now. Can wait til it's needed.

PR #232, opened 2025-02-25
chore: fix windows dev install

lucaseck:
Windows dev install was failing with:
```
ValueError: None is not a valid CPUArch
```
Added `AMD64` as a valid CPU architecture which is what Windows seems to be returning.
Platform.machine() seems to eventually call https://github.com/python/cpython/blob/main/Lib/platform.py#L836 where x86_64 is not a valid option ever for Windows. It looks like there may be other options, but for now I'm just adding AMD64 and can revisit if other Windows machines run into issues.

Also had to implement the fix from https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/178 for the CIs to pass
Windows dev install works
Did a dev install for Houdini 19.5 and verified that the Deadline Cloud node was available

No

No

----


epmog (dismissed):

edwards-aws (dismissed):

edwards-aws (approved):

epmog (approved):

leongdl (commented):
LGTM

PR #183, opened 2024-10-21
fix: hou.fileReferences with parameter expressions

paulgolter:
Filepath parameters can contain expressions referencing top level parameter. The current code is calling `parm.unexpandedString()` which will throw and exception in that case. 

This patch fixes this by using the paths that are returned by `hou.filReferences` as they are already unexpanded.  And to expand them for file / directory check, use `hou.expandString()` instead.

### What was the problem/requirement? (What/Why)

The code that collects external references of the hip file can throw an exception.
Even if filepath parameters contain expressions the current code is trying to call `parm.unexpandedString()` on them which will throw an exception.

You can easily replicate this by using a redshift dome light and load an hdr as a texture. 

(Houdini 20.0.688 / Redshift 2025.0.1)

### What was the solution? (How)
Use the filepaths that `hou.fileReferences()` returns as they are already unexpanded. And to expand them for file / directory check, use `hou.expandString()` instead.

### What is the impact of this change?
Just a minor change in the `_get_scene_asset_references` function and an update of the test case.

### How was this change tested?
Updatet the test code. Ran `hatch run test` and tested submission with houdini hip files.

###  Have you run the unit tests?
yes

### Did you run the "Job Bundle Output Tests"? If not, why not? If so, paste the test results here.

I am not sure what "Job Bundle Output Tests" refers to. But i ran `hatch run test`. This is the output of that:

```
Name                                                                                   Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------------------------------------------------------------
src\deadline\houdini_adaptor\HoudiniAdaptor\__init__.py                                    3      0      0      0   100%
src\deadline\houdini_adaptor\HoudiniAdaptor\adaptor.py                                   219      9     60      7    94%   89, 199->228, 231, 288->exit, 343, 366, 371, 471->474, 497->496, 504-513
src\deadline\houdini_adaptor\HoudiniClient\__init__.py                                     3      0      0      0   100%
src\deadline\houdini_adaptor\HoudiniClient\houdini_client.py                              29      2      4      0    94%   15, 42
----------------------------------------------------------------------------------------------------------------------------------
TOTAL                                                                                    887    250    280     28    68%
Coverage HTML written to dir build/coverage
Coverage XML written to file build/coverage/coverage.xml

Required test coverage of 65.0% reached. Total coverage: 68.12%
=================================================================================== slowest 5 durations ==================================================================================== 
[...]

PR #235, opened 2025-02-25
fix: bug that caused unhandled exceptions during file parsing

lucaseck:
Copied/based on the previous work and bug report of @paulgolter in #183 due to inactivity on that PR

When parsing the files to include for submission we were calling parm.unexpandedString() on each file path. Filepaths in Houdini can contain a static string path or an expression referencing another parm etc. In certain cases you're unable to call unexpandedString() on the file path parms without causing an exception. Specifically RedShift dome lights have been brought up in the previous PR and other reports. RedShift dome lights contain an inner file path parm with keyframes set on it which results in an error like below when unexpandedString() is called:
```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.805/houdini/python3.9libs\hou.py", line 66339, in unexpandedString
    return _hou.Parm_unexpandedString(self)
hou.OperationFailed: The attempted operation failed.
Cannot get unexpanded string for parms with keyframes
```

hou.fileReferences() already returns Houdini's interpretation of the parm as an unexpanded string according to their [docs](https://www.sidefx.com/docs/houdini/hom/hou/fileReferences.html). Instead of calling unexpandedString() again we can just use what hou.fileReferences() has already returned.

I've also removed previous debug code that was added to help diagnose which nodes were causing issues with the unexpandedString() calls as we no longer call that function ourselves here.

* The use of RedShift dome lights specifically no longer causes unhandled exceptions during file parsing
* Other unknown nodes or use cases should also not run into issues from the additional call to unexpandedString()

* In Houdini 19.5 I created a simple scene with just a RedShift dome light and the Deadline Cloud node
* I added a texture to the RedShift dome light
* Clicked parse files and saw the unhandled exception that the prior PR must have been referring to and visible above
* Modified the submitter code as seen in this PR
* Clicked parse files again and no longer had any unhandled exceptions and the texture on the RedShift dome light was successfully added to the list of input filenames

Yes, and they have been updated to reflect the new behaviour 

Added a description to the commit to hopefully make the changelog more clear of why this was changed

Shouldn't be no.
----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
@@ -182,9 +182,6 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
     asset_references = AssetReferences()
     asset_references.input_filenames.add(_get_hip_file())
 
-    # collect references that we could not evaluate so we can surface these later.
-    failure_messages: dict[str, list[tuple[hou.Parm, str]]] = {}
-
     for parm, ref in hou.fileReferences():
         if (
             (not parm)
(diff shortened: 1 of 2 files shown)


src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
```diff
@@ -198,34 +195,13 @@ def _get_scene_asset_references(rop_node: hou.Node) -> AssetReferences:
...
-            failed_references_msg,
-            title="Houdini Asset Parsing",
-            severity=hou.severityType.Warning,
-            details=failed_references_details,
-            details_expanded=True,
-        )
+        # hundreds/thousands. hou.fileReferences() already returns the refs as
+        # the equivalent of calling parm.unexpandedString()
+        # https://www.sidefx.com/docs/houdini/hom/hou/fileReferences.html
+        if os.path.isdir(path):
```
leongdl, line 220: Is the exception that we caught before always not going to be thrown now?

lucaseck: Correct, it was a direct result of us calling `unexpandedString()` which is not called at all anymore in this section. We are just taking the refs given by Houdini and adding them to the lists as is.

leongdl: Got it :) Simplification.

leongdl (approved):

moorec-aws (approved):

PR #244, opened 2025-03-26
feat: Add standalone installer

miabatta:
## What was the problem/requirement? (What/Why)
We want to add a standalone installer for Houdini!

- Add a new XML document defining the individual installer
- Add scripts for building and testing the installer

We can build standalone installers for Houdini :)

Ran the new installer tests on Linux and Windows, also ran existing unit tests to check for regressions

n/a, no submitter/adaptor changes

Added instructions on building & testing the new installer in `DEVELOPMENT.md`.

nope

----


epmog (dismissed):

installer/DeadlineCloudForHoudiniSubmitter.xml
```diff
@@ -0,0 +1,244 @@
...
+                <if>
+                    <conditionRuleList>
+                        <platformTest type="linux"/>
+                        <compareText logic="equals" text="${scope}" value="system"/>
+                    </conditionRuleList>
+                    <actionList>
+                        <!-- InstallBuilder's system-wide Linux implementation does not work (it always writes to the user
+                        environment files like ~/.bashrc and ~/.bash_profile, even as root).
+                        Instead, we cache all environment variables in a global variable, and in the post-installation
+                        steps, we generate a file in /etc/profile.d/ that will get sourced on login. -->
```
epmog: Looks like this was deleted from the postinstallation step, as we're not actually setting it.

If we're not using the function now we can just delete it. Otherwise we should add it back in

epmog (approved):

MarinMaksutaj (approved):

PR #257, opened 2025-04-28
fix: ROP Geometry nodes not being evaluated

### LINKED ISSUE #255: Bug: ROPGeometryOutput SOP Nodes are not respected correctly by Deadline Cloud Submitter
### Describe the bug

When you are referencing a SOP 'ROP Geometry Node' , via a 'fetch' node into an Output Graph, to feed into a Deadline Cloud submitter, then it causes a few issues:

(ROP Geometry nodes are in filecache for example)

- when you 'Parse Files' on the submitter ROP it fails to find any outputs ( so by default it wont save the result in a JA environment)

- the initsim flag ( Initialise Sim) option is not picked up, resulting in a sim task per frame, not 1 task for the entire sim.

### Expected Behaviour

Deadline Cloud's ROP should evaluate ROP Geometry Nodes as it would evaluate Geometry Output Nodes.

### Current Behaviour

Fails to generate outputs by defaults, and sims fail.

### Reproduction Steps

create a ROP Geometry node in sops, fetch it into and Output context, and connect to a Deadline node.

### Environment

At minimum:
1. Operating system Windows 2022
2. Version of Houdini: 20.0.724
3. Version of this package

lucaseck:
Fixes: #255 

ROP geometry nodes were not being evaluated to detect output paths or if they should be run as a single task for simulations

The ROP Geometry node has a different driver type than the regular Geometry nodes so I've added the driver to the list of nodes to check for `initsim` and the mapping for where to look for output files.

* Output files are found
* initsim being selected results in only a single task being created

* Created a scene locally with a ROP Geometry node
* Selected the Initialize Simulation option in the node
* Parsed files and saw that the output directory was not added to the list
* Submitted the job and saw that one task was created for every frame despite initsim being set

Made my changes
* Parsed files and saw that the output directory was added to the list correctly
* Submitted the job and saw that only a single task was created for all frames so initsim was being respects

Also tried using a fetch node like the original bug described and everything still worked correctly.

Yes

Documented in the GitHub issue

No

----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -223,7 +223,7 @@ def _get_render_strategy_for_node(node: hou.Node) -> RenderStrategy:
     render_strategy = RenderStrategy.PARALLEL
 
     if (
-        node.type().nameWithCategory() == "Driver/geometry"
+        node.type().nameWithCategory() in ["Driver/geometry", "Sop/rop_geometry"]
         and node.parm("initsim")
         and node.parm("initsim").eval()
     ):
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/_assets.py
@@ -344,6 +344,7 @@ _NODE_DIR_MAP = {
     "Driver/ris::3.0": _renderman_outputs,  # Renderman
     "Driver/Redshift_ROP": "RS_outputFileNamePrefix",  # Redshift
     "Sop/rop_alembic": "filename",  # ROP Alembic Output
+    "Sop/rop_geometry": "sopoutput",  # ROP Geometry Output
     "Dop/rop_dop": "dopoutput",  # ROP Output Driver
     "Driver/usdrender": _usd_render_outputs,  # USD Render ROP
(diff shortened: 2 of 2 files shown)


epmog (approved):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -223,7 +223,7 @@ def _get_render_strategy_for_node(node: hou.Node) -> RenderStrategy:
     render_strategy = RenderStrategy.PARALLEL
 
     if (
-        node.type().nameWithCategory() == "Driver/geometry"
+        node.type().nameWithCategory() in ["Driver/geometry", "Sop/rop_geometry"]
```
miabatta, line 226: QQ, is this something we'll just have to add to as errors come up? Wondering if we can be a bit more proactive or if it's a one-off error that this will be sufficient for.

miabatta (approved):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
lucaseck: It's possible yes. This is our default list of which nodes we assume should be running simulations and default to creating one task for the entire sim. The workaround for anything that we don't pick up automatically is that users can manually tell our Node to treat other nodes as a "sim" and create only one task for the entire frame range and that's documented [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/docs/user-guide.md#parallel-vs-sequential-rendering)

Issue #258, opened 2025-05-13
Bug: Dev install workflow breaks on Houdini 19.5


miabatta:
### Describe the bug

It seems some requirements added for installer scripts are interfering with dev installations of the Houdini submitter. The script will fail if `--houdini-version` is 19.5 because it uses Python 3.9; 20.0 and 20.5 are unaffected.

### Expected Behaviour

Houdini plugin can register using the dev workflow out-of-the-box.

### Current Behaviour

Following output:
```
ERROR: Ignored the following versions that require a different python version: 8.2.0 Requires-Python >=3.10
ERROR: Could not find a version that satisfies the requirement click==8.2.0 (from versions: 0.1, 0.2, 0.3, 0.4, 0.5, 0.5.1, 0.6, 0.7, 1.0, 1.1, 2.0, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 3.0, 3.1, 3.2, 3.3, 4.0, 4.1, 5.0, 5.1, 6.0, 6.1, 6.2, 6.3, 6.4, 6.6, 6.7, 7.0, 7.1, 7.1.1, 7.1.2, 8.0.0a1, 8.0.0rc1, 8.0.0, 8.0.1, 8.0.2, 8.0.3, 8.0.4, 8.1.0, 8.1.1, 8.1.2, 8.1.3, 8.1.4, 8.1.5, 8.1.6, 8.1.7, 8.1.8)
ERROR: No matching distribution found for click==8.2.0
```
```
Traceback (most recent call last):
  File "<path>\deadline-cloud-for-houdini/scripts/install_dev_submitter.py", line 200, in <module>
    install_submitter_package(args.houdini_version, cpu_arch, local_deps)
  File "<path>\deadline-cloud-for-houdini/scripts/install_dev_submitter.py", line 144, in install_submitter_package
    _build_deps_env(
  File "<path>\deadline-cloud-for-houdini/scripts/install_dev_submitter.py", line 133, in _build_deps_env
    subprocess.run(args, check=True)
  File "<path>\Python\Python312\Lib\subprocess.py", line 573, in run
    raise CalledProcessError(retcode, process.args,
[...]

PR #259, opened 2025-05-13
fix: Add manual path mapping for environment variables

### LINKED ISSUE #245: Bug: $JOB token doesn't work when used in output of a Redshift render
### Describe the bug

When you use $JOB in the output path of a Redshift render, the full path gets added to the Job Attachments and then the adaptor fails to pathmap it.

### Expected Behaviour

$HIP works fine, so I would expect both $JOB and $HIP to have the same behavior

### Current Behaviour

When you hit "Parse Files" the full path gets added to the job attachments. Then when the render happens, the worker looks for that full path, even if it was a path local to the submitting computer. The error log I see specifically is `[Redshift] Output file locked or path not available for writing. Skipping frame. (C:/Users/<username>/Desktop/houdini_testA/houdini/render/renderTest_B_v001.Redshift_ROP1.0001.exr.lock)`

This is especially strange because earlier in the log I see this line:
`Setting HOUDINI_PATHMAP to: {'C:/Users/shop2/Desktop/houdini_testA/houdini': '/sessions/session-1a75cae16848476aa615b3a4c16b6947kwb72iia/assetroot-1d8eff26e400ca1b7663'}` so maybe this issue is also related to https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/228

Manually changing the full path back to $JOB doesn't seem to fix the issue.

### Reproduction Steps

Create a Houdini render with $JOB in the redshift output.

Try to render it.

### Environment

At minimum:
1. Operating system: Submitting from Windows to Linux SMF
2. Version of Houdini: 20.5
3. Version of this package: Latest. We are on SMF

miabatta:
Fixes: #245 

Redshift render nodes using the `$JOB` token will not be pathmapped properly, causing renders to fail. Please see the linked issue for more context.

Re-pathmap environment variables after opening a scene s.t. the `$JOB` token will be properly re-evaluated before rendering to a location that doesn't exist.

This behaviour is only seen with Redshift, possibly because of an error with its pathmapping (i.e., any output path using `HOUDINI_PATHMAP` has been fine so far). This is a workaround that allows Redshift nodes to save files to a pathmapped location.

Ran unit tests.

Rendered a test scene with a `RS_Render` node and its output location set to `$JOB/render/file.exr`; saw that the render succeeded and looks as expected.

Houdini 20.0:
```
test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
[gw0] [100%] PASSED test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter

================================================= slowest 5 durations =================================================
16.79s call     test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
0.02s setup    test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
0.00s teardown test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
================================================= 1 passed in 18.75s ==================================================
```
Houdini 20.5:
```
test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
[gw0] [100%] PASSED test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter

================================================= slowest 5 durations =================================================
15.07s call     test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
0.01s setup    test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
0.00s teardown test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter
================================================= 1 passed in 16.05s ==================================================
```

Couldn't run on 19.5; seems the new installer scripts broke our `hatch run install` script for Houdini 20.5. Cut an issue #258

N/A

nop
nop
----

### DIFF
--- a/src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
+++ b/src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
@@ -36,6 +36,38 @@ class HoudiniHandler:
         self.wedge = None
         self.wegenum = None
 
+    def _perform_path_mapping(self) -> None:
+        """
+        Applies the contents of HOUDINI_PATHMAP to global variables
+        in case they are not automatically re-mapped.
+        """
+
+        # Use the HScript command `setenv` to retrieve all global variables as a tuple:
+        # [0] all environment variables in one string, each with the format "KEY\t= VAL\n"
+        # [1] any errors from running the command
+        # https://www.sidefx.com/docs/houdini/commands/setenv.html
+        set_envs, err = hou.hscript("setenv")
+
+        # Early return on errors in case they are benign
+        if err:
+            print(f"Error retrieving Houdini environment variables: {err}")
+            return
+
+        for line in set_envs.split("\n"):
+            if line.strip():
+                key, val = line.split("\t= ", 1)
+                # Use the `pathmap` command to apply HOUDINI_PATHMAP to each variable
+                # in case it missed mapping.
+                # `-t` (test) applies the path mapping rules to the given path
+                # `-c` (compact) prints out just the re-mapped path (with a newline we must strip)
+                mapped_path, err = hou.hscript(f"pathmap -t '{val}' -c")
+                if not err and mapped_path:
+                    hou.putenv(key, mapped_path.strip())
+                    print(f"Set {key} to {mapped_path.strip()}")
+
+        # Reload the env variables
+        hou.hscript("varchange")
+
     def set_node_settings(self, node):
         # this is a place holder function
         # TODO remove after common node library implemented
@@ -204,3 +236,6 @@ class HoudiniHandler:
             hou.hipFile.load(scene_path)
         except hou.LoadWarning as e:
             print(e)
+            return
+
+        self._perform_path_mapping()
--- a/src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
+++ b/src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
@@ -36,34 +36,28 @@ class HoudiniHandler:
         self.wedge = None
         self.wegenum = None
 
-    def _perform_path_mapping(self) -> None:
+    def _path_map_envs(self) -> None:
         """
         Applies the contents of HOUDINI_PATHMAP to global variables
-        in case they are not automatically re-mapped.
+        that are not automatically re-mapped.
         """
 
-        # Use the HScript command `setenv` to retrieve all global variables as a tuple:
-        # [0] all environment variables in one string, each with the format "KEY\t= VAL\n"
-        # [1] any errors from running the command
-        # https://www.sidefx.com/docs/houdini/commands/setenv.html
-        set_envs, err = hou.hscript("setenv")
+        # We know that $JOB and $POSE are not pathmapped automatically,
+        # because they don't change when a scene file moves:
+        # https://www.sidefx.com/docs/houdini/basics/project.html#tips-and-notes
+        envs_to_remap: list[str] = ["JOB", "POSE"]
 
-        # Early return on errors in case they are benign
-        if err:
-            print(f"Error retrieving Houdini environment variables: {err}")
-            return
-
-        for line in set_envs.split("\n"):
-            if line.strip():
-                key, val = line.split("\t= ", 1)
-                # Use the `pathmap` command to apply HOUDINI_PATHMAP to each variable
-                # in case it missed mapping.
-                # `-t` (test) applies the path mapping rules to the given path
-                # `-c` (compact) prints out just the re-mapped path (with a newline we must strip)
-                mapped_path, err = hou.hscript(f"pathmap -t '{val}' -c")
-                if not err and mapped_path:
-                    hou.putenv(key, mapped_path.strip())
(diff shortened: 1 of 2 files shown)


src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
```diff
@@ -36,6 +36,38 @@ def __init__(self) -> None:
...
+        set_envs, err = hou.hscript("setenv")
+
+        # Early return on errors in case they are benign
+        if err:
+            print(f"Error retrieving Houdini environment variables: {err}")
+            return
+
+        for line in set_envs.split("\n"):
+            if line.strip():
+                key, val = line.split("\t= ", 1)
```
AWS-Samuel: Is it possible to have something funky like

```
ARG1\t="multiline\nstring"
ARG2\t="env\t=var"
```

Or does houdini limit what characters can be in the environment?

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
```diff
@@ -36,6 +36,38 @@ def __init__(self) -> None:
...
+                # Use the `pathmap` command to apply HOUDINI_PATHMAP to each variable
+                # in case it missed mapping.
+                # `-t` (test) applies the path mapping rules to the given path
+                # `-c` (compact) prints out just the re-mapped path (with a newline we must strip)
```
AWS-Samuel: These comments are super helpful!

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
```diff
@@ -36,6 +36,38 @@ def __init__(self) -> None:
...
+                mapped_path, err = hou.hscript(f"pathmap -t '{val}' -c")
```
AWS-Samuel: Is there any way to distinguish between a path env var and a not path env var? My concern is that this could end up pathmapping things which shouldn't be modified in the case of generic pathmapping rules. 

E.g. `CAMERA\t=defaultCamera1` becomes `CAMERA\t=$SESSION_DIR/defaultCamera1`

miabatta: Thought about using `Path.exists()` for this purpose but it's not really applicable... though worth noting that we overwrite `HOUDINI_PATHMAP` in the adaptor anyway ([`adaptor.py` lines 340-353](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py#L340-L353)) so we control the input here.

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
miabatta: From looking at the docs it looks like it's not _not_ possible 🤔 There's some tricky delimiting no matter how we go about this. The alternative is to scope mapping down to variables we know need it (`$JOB` and `$POSE`), worst-case scenario we can add more variables as we discover them.

Edit: *assuming that there isn't a better way of fetching all env variables (e.g., a `hou` function). I looked for them while writing the change but am doing a second pass of the docs

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
```diff
@@ -36,6 +36,32 @@ def __init__(self) -> None:
...
+        that are not automatically re-mapped.
+        """
+
+        # We know that $JOB and $POSE are not pathmapped automatically,
+        # because they don't change when a scene file moves:
+        # https://www.sidefx.com/docs/houdini/basics/project.html#tips-and-notes
+        envs_to_remap: list[str] = ["JOB", "POSE"]
+
+        for env in envs_to_remap:
+            print(f"Mapping {env} ({hou.getenv(env)})")
```
lucaseck, line 51: Does getenv play nicely if the env var doesn't exist in Houdini? I know we are checking JOB and POSE, but not sure if someone can make it so those aren't set at all and what happens in that case.

miabatta: Thanks for catching- this will return None if the variable doesn't exist. The `pathmap` call after it, in that case, will just return an empty string and nothing will happen.

Edit: reference here https://www.sidefx.com/docs/houdini/hom/hou/getenv.html

lucaseck: Super nit: I could see printing `Mapping JOB None` confusing people if that does happen and appears in their logs. Could short circuit things and just check if env is set first thing and break if not to just skip the rest.

lucaseck (approved):

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
AWS-Samuel: I think it would just show up as `Mapping Job ` since `None` casted to a string is `''`

AWS-Samuel (approved):

PR #261, opened 2025-05-20
feat: add Redshift pathmapping support

### LINKED ISSUE #228: Feature request: Add pathmapping for Redshift renders
### Describe the problem

Redshift nodes don't seem to respect HOUDINI_PATHMAP which is how the adaptor currently does pathmapping.

### Proposed Solution

For Redshift renders, we'll need to set REDSHIFT_PATHOVERRIDE_STRING in addition to HOUDINI_PATHMAP. Should be a pretty simple fix.
docs on Redshift environment variables: https://help.maxon.net/r3d/houdini/en-us/Content/html/Redshift+Environment+Variables.html

### Example Use Cases

Anyone looking to use job attachments with Redshift Houdini renders would benefit from this feature.

lucaseck:
Fixes: #228  

HOUDINI_PATHMAP appears to work on Redshift nodes directly in the Houdini scene, but for any references inside Redshift proxy files where the relative paths are no longer correct there is no built-in support to re-pathmap them. Note: Redshift proxies do contain both the relative and absolute paths of any external file references, but there can be situations where the relative paths are no longer correct and the absolute paths need to be remapped.

In addition to HOUDINI_PATHMAP the adaptor will now also collect any pathmapping rules from the service or set by REDSHIFT_PATHOVERRIDE_STRING and REDSHIFT_PATHOVERRIDE_FILE and combine them into a new temporary file that REDSHIFT_PATHOVERRIDE_FILE is then set to for Redshift to pathmap any references it needs to. Verified that the override file can have all of the rules on the same line.

The Redshift specific pathmapping variable is set.

* Ran all existing unit tests and added new ones related to pathmapping including many scenarios with the new Redshift logic
* Did a submission with our current adaptor available in SMF with a scene that references a Redshift proxy file and verified that I got missing texture errors and the texture was missing from the output
* Did a submission with a dev build of the adapator with the Redshift pathmapping changes and the same scene as above. Verified that I saw the logging messages of REDSHIFT_PATHOVERRIDE_FILE being set and that during rendering there were no errors for missing textures. The output showed the texture from the proxy file as well.
* Ran a real test with a queue env setting REDSHIFT_PATHOVERRIDE_STRING and FILE to confirm that all rules are being picked up and set in the new file.

Yes

No

No

----

### DIFF
--- a/src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
+++ b/src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
@@ -8,6 +8,7 @@ import re
 import sys
 import threading
 import time
+from pathlib import Path
 from functools import wraps
 from typing import Callable
 
@@ -19,6 +20,7 @@ from openjd.adaptor_runtime.process import LoggingSubprocess
 from openjd.adaptor_runtime.app_handlers import RegexCallback, RegexHandler
 from openjd.adaptor_runtime.application_ipc import ActionsQueue, AdaptorServer
 from openjd.adaptor_runtime_client import Action
+from openjd.adaptor_runtime._utils import secure_open
 
 from .._version import version as adaptor_version
 
@@ -337,12 +339,9 @@ class HoudiniAdaptor(Adaptor[AdaptorConfiguration]):
         else:
             os.environ["PYTHONPATH"] = python_path_addition
 
-        # If there are path mapping rules, set the houdini environment variable to enable them
-        houdini_pathmap = self._get_houdini_pathmap()
-        if houdini_pathmap:
-            os.environ["HOUDINI_PATHMAP"] = houdini_pathmap
-
-        _logger.info("Setting HOUDINI_PATHMAP to: {}".format(houdini_pathmap))
+        # Set any path mapping rules
+        self._set_houdini_pathmap()
+        self._set_redshift_pathmap()
 
         houdini_client_path = self._get_houdini_client_path()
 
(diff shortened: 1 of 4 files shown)


src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
```diff
@@ -352,24 +351,90 @@ def _start_houdini_client(self) -> None:
...
-        for rule in self._path_mapping_rules:
-            path_mapping_rules[rule.source_path.replace("\\", "/")] = rule.destination_path.replace(
-                "\\", "/"
+        redshift_rules: set[tuple[str, str]] = {
+            (rule.source_path.replace("\\", "/"), rule.destination_path.replace("\\", "/"))
+            for rule in self.path_mapping_rules
+        }
+
+        # "C:/Users/AUser/Directory" "/sessions/session-abcd/"
+        redshift_rule_regex = re.compile(r"\"([^\"]*)\"\s+\"([^\"]*)\"")
```
erico-aws: I think the comment here shows an example of what this regex captures, but do you mind adding a note that explains what it's capturing?

lucaseck: Added!

erico-aws (dismissed):

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
```diff
@@ -19,6 +20,7 @@
 from openjd.adaptor_runtime.app_handlers import RegexCallback, RegexHandler
 from openjd.adaptor_runtime.application_ipc import ActionsQueue, AdaptorServer
 from openjd.adaptor_runtime_client import Action
+from openjd.adaptor_runtime._utils import secure_open
```
epmog, line 23: importing this function from a private module tells us we're doing something wrong. Why do we need `secure_open` vs. just a regular `open`?

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
```diff
@@ -352,24 +351,101 @@ def _start_houdini_client(self) -> None:
...
+            except Exception as e:
+                _logger.warning(
+                    f"Error reading the REDSHIFT_PATHOVERRIDE_FILE at {existing_rule_file_path}: {str(e)}"
+                )
 
-        if path_mapping_rules:
-            return str(path_mapping_rules)
-        return ""
+        # Collect any additional rules if REDSHIFT_PATHOVERRIDE_STRING is already set
+        if existing_rule_str := os.getenv("REDSHIFT_PATHOVERRIDE_STRING"):
```
epmog: should we only do this if file isn't set? To match redshift behaviour?

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
```diff
@@ -352,24 +351,101 @@ def _start_houdini_client(self) -> None:
...
+        redshift_rules: set[tuple[str, str]] = {
+            (rule.source_path.replace("\\", "/"), rule.destination_path.replace("\\", "/"))
+            for rule in self.path_mapping_rules
+        }
+
+        # Regex to capture two groups. Each group consists of any non double quote
+        # characters between two sets of double quotes separated by whitespace.
+        # For the example rule "C:/Users/AUser/Directory" "/sessions/session-abcd/"
+        # it will capture C:/Users/AUser/Directory and /sessions/session-abcd/
+        redshift_rule_regex = re.compile(r"\"([^\"]*)\"\s+\"([^\"]*)\"")
```
epmog: Alternative: Can we not just copy the lines from the variables/file wholesale and just add a newline after we add our rules? Would remove all of the regex/parsing and leave everything prior intact.

----------

Otherwise, it'd be easier to read the regex with single-quoted string to reduce the characters in it. Can there be double quotes within the double-quoted paths? 

Can probably simplify to something like this
```python
re.compile(r'^"(.+)"\s+"(.+)"$')
```
We're now anchored at the beginning and end, there has to be a value between the quotes.

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
lucaseck: Good question, not entirely familiar with how that function came to be, but I believe copied it from the Maya adaptor which does the same import and uses it to write its pathmapping file for arnold: https://github.com/aws-deadline/deadline-cloud-for-maya/blob/c7bdb585d9de02667ec924fed6e8c2f7ea26b598/src/deadline/maya_adaptor/MayaAdaptor/adaptor.py#L590

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
lucaseck: For the string we can't copy it wholesale, we need to parse the pairs in there since the file needs to be one rule per line is my understanding. I'd have to test that manually to see if it can all be one line or not, since their docs aren't clear,  but the [example](https://help.maxon.net/r3d/maya/en-us/Content/html/Redshift+Environment+Variables.html) shows separate lines.

As for double quotes being within the paths I haven't tested it, but I doubt that's allowed. `"` is not a valid path character on Windows at least and can't be used in filepaths where the whole point of this variable is to be able to map paths between OSes etc.

I can change to wrap it in single quotes since I agree that helps with readability, but the anchoring in that example regex: `re.compile(r'^"(.+)"\s+"(.+)"$')` I don't believe works when you have multiple rules in a row in the string representation of the override. It will group all of the initial rules in group1 and just the last into a group2.

e.g. `"C:/Users/AUser/Directory" "/sessions/session-abcd/" "C:/Users/AUser/Directory1" "/sessions/session-abcd1/"` results in group 1 being `C:/Users/AUser/Directory" "/sessions/session-abcd/" "C:/Users/AUser/Directory1` and group 2 being `/sessions/session-abcd1/`

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
lucaseck: Good thought, I can't think of a reason not to align with that since that's what the regular behaviour is.

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
epmog: Does the path_override_file also handle everything on one line? Might therefore still make it all moot.

-----

Yea, based on their example, I was presuming that there was 1 rule per line even for the string override. If they just have groupings of paths within quotes, then we can remove the anchoring. If there are no double quotes within the source or dest, or if it's unlikely to be, then you can do the character set exclusion or make the quantifier lazy. Lazy (as-opposed to the default greedy) example: `"(.+?)"\s+"(.+?)"`

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
lucaseck: Spoke offline and got consensus that there's a reason this is being done and while not ideal it should stay this way for now.

epmog (approved):

erico-aws (approved):

src/deadline/houdini_adaptor/HoudiniAdaptor/adaptor.py
lucaseck: Ended up testing this and yes, the override file can have multiple rules on a single line and combined with more rules on additional lines. PR has been updated to simplify that process.

PR #303, opened 2025-09-11
build: remove installer images from git lfs

justinsaws:
The installer images being in git lfs causes the installer builds to break.

Recently, the installer .pngs were moved to git lfs but this breaks the installer building process because our automated builds don't do a git lfs checkout when they build the code.

Revert this change.

Installers work again.

`hatch build && hatch run test --all`

----


miabatta (dismissed):
Do you know why the integ test images aren't showing up in the diff as well? Just curious

erico-aws (approved):

miabatta (approved):

PR #309, opened 2025-10-03
fix: aws credential error message on node creation is not actionable

### LINKED ISSUE #134: Bug: expired credentials message isn't actionable
### Expected Behaviour

The message should be more actionable. For example, if the user is using Deadline Cloud monitor credentials, we could open the Deadline Cloud monitor and ask the user to log in.

### Current Behaviour

Currently, when the AWS Deadline Cloud credentials are expired, a message "AWS Deadline Cloud credentials are expired." will appear in the Houdini Console. While this message is informative, it isn't clear what the user should do to resolve the issue.

### Reproduction Steps

Pre-requesites:
* Deadline Cloud credentials are expired

Reproduction Steps:
1. Install Deadline Cloud for Houdini submitter
2. Open the `out` network
3. Create a Deadline Cloud ROP
4. An error message appears "AWS Deadline Cloud credentials are expired."

### Code Snippet

The error message comes from [here](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/c69ea0bdb40190d286d4e36ccfefc5efb0f1fafd/src/deadline/houdini_submitter/otls/deadline_cloud.hda/Driver_1deadline__cloud/OnCreated#L16).

There is similar functionality in the Deadline Cloud for Unreal Engine [here](https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/blob/bf6ae92a303a2f5b57fafef7a8a34939257a99bf/src/unreal_plugin/Content/Python/settings.py#L358) that makes the expired credentials message more actionable. Perhaps this could be in a shared function in the [Deadline Cloud client library](https://github.com/aws-deadline/deadline-cloud/tree/34216d7e3667fec894415b1a0341f80639fb4521/src/deadline/client)?

epmog:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/134

We had a custom error message that didn't tell users what to do when their credentials were expired.

We now just pass the execption string which provides instructions for users to refresh their credentials.

At the same time, I've added a warning dialog for settings/login if the credentials aren't refreshed that prints the same information, but optionally includes the stack trace about the actual error

image

Customers have better instructions on how to fix their set-up

Deployed it locally

N/A

N/A

N/A

----

### DIFF
--- a/src/deadline/houdini_submitter/otls/deadline_cloud.hda/Driver_1deadline__cloud/OnCreated
+++ b/src/deadline/houdini_submitter/otls/deadline_cloud.hda/Driver_1deadline__cloud/OnCreated
@@ -3,5 +3,5 @@ from botocore.exceptions import CredentialRetrievalError
 try:
     node = kwargs['node']
     node.hdaModule().update_queue_parameters_callback(kwargs)
-except CredentialRetrievalError:
-    print('AWS Deadline Cloud credentials are expired.')
+except CredentialRetrievalError as e:
+    print(e)
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -701,14 +701,30 @@ def settings_callback(kwargs):
     node = kwargs["node"]
     _show_farm_and_queue_as_refreshing(node)
     DeadlineConfigDialog.configure_settings(parent=hou.qt.mainWindow())
-    _apply_farm_and_queue_settings(node)
+    try:
+        _apply_farm_and_queue_settings(node)
+    except Exception as exc:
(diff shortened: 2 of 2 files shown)


src/deadline/houdini_submitter/otls/deadline_cloud.hda/Driver_1deadline__cloud/OnCreated
```diff
@@ -1,7 +1,8 @@
+import hou
```
erico-aws: What's this import for?

epmog: leftover from the initial solution :)

jusiskin (approved):

erico-aws (approved):

PR #312, opened 2025-10-21
docs: Add user guide infrastructure

miabatta:
We have a global integrations landing page! Let's add some information about the Houdini submitter.

Added a mkdocs site with information about installing and using the submitter.

https://github.com/user-attachments/assets/e85b36da-61c6-4fdf-b61e-c31ce8f25397

Better customer-facing docs

ran `hatch run docs:serve` to test out the site locally

n/a

I guess you could say that

no

----


docs/user_guide/installation.md
```diff
@@ -0,0 +1,26 @@
+# Installation
+
+To install the AWS Deadline Cloud for Houdini submitter, you will need:
+
+* A Windows, macOS, or Linux workstation
+* Houdini 19.5 or later
```
lucaseck: Should we include the upper bound? It's annoying to have to update each time, but we have given the impression before that versions of Python etc are supported with things when they are untested by not putting upper bounds.

docs/user_guide/installation.md
```diff
@@ -0,0 +1,26 @@
...
+**To install the submitter**
+
+1. Download the [Deadline Cloud submitter installer](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html).
+1. Run the installer.
+    - When prompted, select each version of Houini you want to use the submitter with.
+1. Launch Houdini.
+
+The Deadline Cloud submitter should be automatically available as a render operator (ROP) node.
+
+> The submitter installer is available for Windows, macOS, and Linux. See the [developer README](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/README.md) for manual installation instructions.
```
lucaseck, line 17: nit: can we specify ARM macOS?

docs/user_guide/submitter/index.md
```diff
@@ -0,0 +1,8 @@
+# Using the Deadline Cloud for Houdini Submitter
+
+The Deadline Cloud for Houdini submitter is a node that accepts a render operator (ROP) node as input. You can configure and submit your job through this node.
```
lucaseck: To me this gives the impression that only a single ROP node can be used as input. We do walk the connected graph to create multiple steps in the job by default, one for each ROP node in the tree.

docs/user_guide/submitter/using-submitter.md
```diff
@@ -0,0 +1,42 @@
...
+- A profile to submit to Deadline Cloud with
+- A Deadline Cloud farm and queue to submit to
+
+## Submit a job
+
+**To submit a job from Houdini to Deadline Cloud**
+
+1. In the Network Editor, select the **/out** network.
+1. Open the context menu (right-click or press **Tab**) and search for `deadline` to create a Deadline Cloud node.
+1. Connect the output of a ROP to the input of the Deadline Cloud node.
```
lucaseck, line 14: Same comment as above, might be worth calling out it doesn't have to just be a single node. We walk the tree and I think by default separate steps for each rop node in the graph

docs/user_guide/submitter/using-submitter.md
```diff
@@ -0,0 +1,42 @@
...
+1. Select the Deadline Cloud node.
+1. Use the options in the node editor to configure your job. See [Houdini-specific Settings](#houdini-specific-settings) for information about what each option does.
+1. Choose **Login** to log into your AWS profile if not already logged in.
```
lucaseck: nit: Can we specify that Login is for your DCM Profile, if you're just using a named AWS profile the login will just throw a warning.

docs/user_guide/submitter/using-submitter.md
```diff
@@ -0,0 +1,42 @@
...
+1. (Optional) To export a job's associated files to your job history directory without submitting it, choose **Export Bundle**.
+    - A _job bundle_ is a group of files that defines a job. For more information, see [Open Job Description templates for Deadline Cloud](https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/build-job-bundle.html).
+1. Choose **Submit** to send your job to Deadline Cloud.
+
+## Houdini-specific Settings
+
+The **Job-specific settings** tab of the Deadline Cloud node provides options specific to Houdini jobs.
+
+* _Submit Dependencies as Separate Steps_ - Split the ROP graph into separate rendering steps for easier monitoring and debugging. When enabled, each connected render node becomes its own step in the job.
+* _Include Adaptor Wheels_ - Enable custom builds of the adaptor that change rendering behavior. When enabled, you can specify a directory containing adaptor wheels.
```
lucaseck: Can you link to the instructions for how to build the wheels. https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/scripts/build_wheels.sh

crowecawcaw (commented):
Looks like this PR has some good feedback. LGTM otherwise!

docs/user_guide/submitter/using-submitter.md
miabatta: This was also feedback on the Blender guide that I forgot to copy over. I will just remove this message as I did there

docs/user_guide/submitter/using-submitter.md
miabatta: Yep, I think a brief explanation of what a "wheel" is might be a good idea as well

lucaseck (approved):

crowecawcaw (approved):

PR #327, opened 2025-12-02
ci: add runner setup script for integration tests

moorec-aws:
We want to add houdini integration tests for various houdini versions as part of our CI/CD.

1. Updated the dev installer to support multiple versions of houdini installations on a single system.
2. Added a setup_runner.py script that installs houdini and sets up the submitter for each houdini version.
3. Updated the test_runners.py in the integ test code to run openjd cli per step to workaround a known bug where session ids are not unique when using the cli.
4. Updated hatch.toml with a new environment called `integ-ci` that utilized a matrix of the different houdini versions. It has scripts that setup up the runner and runs the tests for each houdini version.
5. Loosened image comparison to error if there are pixels differences of 2 or greater in .5% of total pixels. This is to account for render differences between different platforms but still strict enough to catch noticeable differences.

Houdini Integration Tests can be added to CI/CD

https://github.com/moorec-aws/deadline-cloud-for-houdini/actions/runs/19876349218

Yes

No.

----


.github/workflows/integration_test.yml
```diff
@@ -5,35 +5,35 @@ on:
 
 jobs:  
   MainlineLinuxIntegrationTest:
-    name: Linux Integration Tests
+    name: Linux Integration Test
     permissions:
       id-token: write
       contents: read
-    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
```
moorec-aws: Will change before merging.

.github/workflows/integration_test.yml
```diff
@@ -5,35 +5,35 @@ on:
...
       environment: mainline
       os: linux
   MainlineWindowsIntegrationTest:
-    name: Windows Integration Tests
+    name: Windows Integration Test
     permissions:
       id-token: write
       contents: read
-    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
```
moorec-aws: Will change before merging.

.github/workflows/integration_test.yml
```diff
@@ -5,35 +5,35 @@ on:
...
       environment: mainline
       os: windows
   MainlineMacOsIntegrationTest:
-    name: MacOS Integration Tests
+    name: MacOS Integration Test
     permissions:
       id-token: write
       contents: read
-    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
```
moorec-aws: Will change before merging.

hatch.toml
```diff
@@ -43,6 +43,22 @@ test_adaptors = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m adap
 [envs.e2e.scripts]
 test = "pytest --no-cov {args:test/*/e2e}"
 
+[envs.integ-ci]
+pre-install-commands = [
+  "pip install -r requirements-integ-testing.txt",
+  "pip install -r requirements-testing.txt"
+]
```
jusiskin, line 48: Same suggestion as aws-deadline/deadline-cloud-for-blender#296.


```suggestion
pre-install-commands = [
  "pip install -r requirements-integ-testing.txt -r requirements-testing.txt",
]
```

scripts/install_dev_submitter.py
```diff
@@ -93,33 +97,27 @@ def _resolve_dependencies(local_deps: list[Path]) -> dict[str, str]:
...
-    resolved_dependencies_dict = _resolve_dependencies(local_deps)
-    resolved_dependencies = [
-        f"{dep_name}=={resolved_version}"
-        for dep_name, resolved_version in resolved_dependencies_dict.items()
-    ]
+    resolved_dependencies = _resolve_dependencies(local_deps, python_version)
+
+    # Add pywin32 for Windows to ensure win32file is available
+    if platform.system() == "Windows":
+        resolved_dependencies.append("pywin32")
```
jusiskin: This seems like something that could get out of sync. Do you know why this is necessary? Is this a limitation or bug in the  `_resolve_dependencies` function called above?

pipeline/setup-runner.py
```diff
@@ -0,0 +1,347 @@
...
+    "21.0.440": {
+        "linux": "a87451f9146d52051a9ba142d535936638351526a48cba0c6156a221f3be58e6",
+        "windows": "a78e468e99d1be3476b46062e3a043ecc435751b9a9f92e39bee414ace8ce59e",
+        "macos": "3fc918428b22b3c32704d1163a6b1b08723ece71e5e0172e941167549d4d5b25",
+    },
+}
+
+
+def run(cmd, check=True):
+    print(f"Running: {cmd if isinstance(cmd, str) else ' '.join(cmd)}")
```
jusiskin: nit: could use [`shlex.join(cmd)`](https://docs.python.org/3/library/shlex.html#shlex.join) for a more correct shell command being logged.

pipeline/setup-runner.py
```diff
@@ -0,0 +1,347 @@
...
+            houdini_dir.mkdir(parents=True, exist_ok=True)
+            run(["chmod", "777", str(houdini_dir)])
+
+            run(
+                [
+                    "bash",
+                    "-c",
+                    f'cd {install_dir} && echo "y" | ./houdini.install --auto-install --install-houdini --no-install-license --no-install-menus --no-install-bin-symlink --install-hfs-symlink --no-install-hqueue-server --no-root-check --accept-EULA 2021-10-13 {houdini_dir}',
+                ]
+            )
```
jusiskin, line 182: nitL we could avoid a shell command here if we made the `stdin` and `cwd` arguments of [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) available in the `run()` wrapper function.

Something like:

```py
# Change the run function
def run(..., stdin=None, cwd=None):
    # ...
    result = subprocess.run(..., stdin=stdin, cwd=cwd)
    # ...


# and then here...
run(
    [
        str(install_dir / "houdini.install"),
        # ...
    ],
    stdin="y".encode("utf8"),
    cwd=install_dir,
)
```

pipeline/setup-runner.py
```diff
@@ -0,0 +1,347 @@
...
+            print(f"Hatch install stderr: {result.stderr}")
+        if result.returncode != 0:
+            print(f"ERROR: Hatch install failed for Houdini {version}")
+            sys.exit(result.returncode)
+
+        # Install pywin32 into Houdini's Python
+        houdini_dir = Path(f"C:/Tools/houdini-{version}")
+
+        # Map Houdini versions to Python versions
+        if version.startswith("19.5"):
```
jusiskin: nit: maybe we should add a trailing `.`


```suggestion
        if version.startswith("19.5."):
```

same for other conditions below

pipeline/setup-runner.py
```diff
@@ -0,0 +1,347 @@
...
+            python_version = "3.9"
+        elif version.startswith("20.0"):
+            python_version = "3.10"
+        elif version.startswith(("20.5", "21.0")):
+            python_version = "3.11"
+        else:
+            python_version = "3.11"
```
jusiskin: Do we want this or maybe raise [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError)?

pipeline/setup-runner.py
```diff
@@ -0,0 +1,347 @@
...
+                "pip",
+                "install",
+                "pywin32<311,>=307",
+                f"--python-version={python_version}",
+                "--only-binary=:all:",
+                f"--target={site_packages}",
+                "--no-cache-dir",
+                "--force-reinstall",
+            ]
+        )
```
jusiskin: Same as my other feedback. Is there any way to make this dynamic and not have to be maintained in two places?

pipeline/setup-runner.py
```diff
@@ -0,0 +1,347 @@
...
+                time.sleep(1)
+                if houdini_marker.exists():
+                    break
+            continue
+
+        lock_file.touch()
+        try:
+            print(f"Installing Houdini {version}...")
+            # Houdini 19.5 uses gcc9.3, others use gcc11.2
+            gcc_version = "gcc9.3" if version.startswith("19.5") else "gcc11.2"
```
jusiskin: For more streamlined maintainability, I wonder if this could be maintained in a single data structure at the top of the file. What do you think?

pipeline/setup-runner.py
```diff
@@ -0,0 +1,347 @@
...
+        houdini_dir = Path(f"C:/Tools/houdini-{version}")
+        houdini_marker = houdini_dir / ".installed"
+
+        if houdini_marker.exists():
+            print(f"Houdini {version} already installed")
+            continue
+
+        print(f"Installing Houdini {version}...")
+        # Houdini 19.5 uses vc142, others use vc143
+        vc_version = "vc142" if version.startswith("19.5") else "vc143"
```
jusiskin: Same feedback as Linux. Can we maintain this in a single data structure at the top of the file?

jusiskin (changes requested):

pipeline/setup-runner.py
moorec-aws: This is the only place this is located. Fulfilling the requirement specified here: https://github.com/moorec-aws/deadline-cloud-for-houdini/blob/mainline/DEVELOPMENT.md#integration-tests. If I don't pin pywin32 the install will fail due to deadline-cloud pinning the version

scripts/install_dev_submitter.py
moorec-aws: This is left over from when I was troubleshooting houdini not being able to find pywin32.

hatch.toml
moorec-aws: Done

scripts/install_dev_submitter.py
moorec-aws: Removed

pipeline/setup-runner.py
moorec-aws: Done

pipeline/setup-runner.py
moorec-aws: Done

pipeline/setup-runner.py
moorec-aws: Added data structure for versions.

pipeline/setup-runner.py
moorec-aws: No longer required.

pipeline/setup-runner.py
moorec-aws: Added data structure for versions.

pipeline/setup-runner.py
moorec-aws: Added data structure for versions.

pipeline/setup-runner.py
jusiskin: We discussed offline and we decided to create a `requirements-<NAME>.txt` file that is references by both the manual integration test instructions in `DEVELOPMENT.md` and in this script.

moorec-aws: Updated to use a requirements file.

jusiskin (dismissed):

DEVELOPMENT.md
```diff
@@ -141,13 +141,12 @@ To run integration tests:
 4. Set the environment variable `HOUDINI_VERSION` to the version of Houdini you want to test against. For example:
    * On Linux: `export HOUDINI_VERSION='20.5.487'`
    * On Windows (powershell): `$Env:HOUDINI_VERSION='20.5.487'`
-   * On Mac: `export HOUDINI_VERSION='20.5.487'`
```
miabatta, line 144: Have you rebased? We just merged this line in #328

miabatta (commented):
Just the one comment otherwise LGTM 👍

DEVELOPMENT.md
moorec-aws: Sure did. Looks like I accidentally removed it sorting out a merge conflict. Nice catch!

moorec-aws: Added back

miabatta (approved):

jusiskin (approved):

PR #329, opened 2025-12-08
ci: add integration tests to merges and release process

moorec-aws:
Integration tests were not running as part of the release process for the Blender DCC adaptor. This meant releases could be published without validating that the integration tests pass on all supported platforms (Linux, Windows, macOS).

We also want integration tests to run when we merge to mainline

Added three integration test jobs to the release_publish.yml workflow:

* LinuxIntegrationTest
* WindowsIntegrationTest
* MacOsIntegrationTest

These jobs run in parallel after the release is tagged and use the reusable integration test workflow from aws-deadline/.github. The PreRelease job now depends on all integration tests passing before proceeding with the release.

Set `integration_tests.yml` to run on push to mainline and added options for running on different branches or tags

* Releases will now be blocked if integration tests fail on any platform, ensuring higher quality releases and catching platform-specific issues before they reach production.

This change modifies the CI/CD workflow itself. Testing will occur when the next release is triggered, at which point the integration tests will run automatically as part of the release process.

**Please run the integration tests and paste the results below.**
N/A - This change adds integration tests to the release workflow. The tests will run automatically on the next release.

**If installer/ was modified or a file was added/removed from src/, then update the installer tests and post the test results below.**
N/A - No changes to installer or src files.

N/A - This is a CI/CD workflow change that doesn't affect user-facing functionality or APIs.

No.

----

### DIFF
--- a/.github/workflows/release_publish.yml
+++ b/.github/workflows/release_publish.yml
@@ -29,13 +29,36 @@ jobs:
 
   UnitTests:
     needs: [TagRelease]
-    name: Unit and Integration Tests
+    name: Unit Tests
     uses: ./.github/workflows/code_quality.yml
     with:
       tag: ${{ needs.TagRelease.outputs.tag }}
 
+  IntegrationTests:
+    needs: [TagRelease, UnitTests]
+    name: ${{ matrix.name}} Integration Tests
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: linux
+            name: Linux
+          - os: windows
+            name: Windows
+          - os: macos
+            name: MacOS
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    secrets: inherit
+    with:
+      repository: ${{ github.event.repository.name }}
+      tag: ${{ needs.TagRelease.outputs.tag }}
+      os: ${{ matrix.os }}
+
   PreRelease:
-      needs: [TagRelease, UnitTests]
+      needs: [TagRelease, UnitTests, IntegrationTests]
       uses: aws-deadline/.github/.github/workflows/reusable_prerelease.yml@mainline
       permissions:
         id-token: write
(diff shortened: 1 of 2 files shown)


jusiskin (dismissed):

.github/workflows/release_publish.yml
```diff
@@ -34,8 +34,47 @@ jobs:
...
+    name: MacOS Integration Tests
+    permissions:
+      id-token: write
+      contents: read
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    secrets: inherit
+    with:
+      repository: ${{ github.event.repository.name }}
+      tag: ${{ needs.TagRelease.outputs.tag }}
+      os: macos
```
erico-aws: Does it make sense to change these jobs to be a matrix job on oses instead of doing 3 separate ones?

moorec-aws: Possibly, I can evaluate that and retest to see what it looks like. Lets get this merged tho so we can have these running in our CI asap.

moorec-aws: Updated. I was able to get this working and retested.

moorec-aws: Actually Im keeping this as is. When a job within a matrix fails, the whole strategy fails, meaning if Windows Fails Linux and MacOS get canceled.

jusiskin: >    When a job within a matrix fails, the whole strategy fails, meaning if Windows Fails Linux and MacOS get canceled.

Looks like this is something you can configure:

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategyfail-fast

jusiskin (dismissed):

.github/workflows/release_publish.yml
moorec-aws: ah yeah. That works. rev up to include this.

erico-aws (approved):

jusiskin (approved):

PR #337, opened 2026-02-05
fix: handle AccessDeniedException when user lacks GetFarm permission

### LINKED ISSUE #336: Bug: Users that only have queue permissions get errors on getFarm calls
### Describe the bug

Users with queue only permissions will see permission errors on attempts to populate the farm name and will be unable to pick queues.  

### Expected Behaviour

Users with queue only permissions should see a farm ID and still be able to select a queue to submit to.

### Current Behaviour

Many errors and no farm/queue displayed. It's possible to submit by editing the config.

### Reproduction Steps

Have a user without farm membership that's only on a queue or queues.

### Environment

At minimum:
1. Operating system: Reproduced on Mac OS
2. Version of Houdini: 21 but likely unrelated to Houdini versions
3. Version of this package: mainline
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? Tested against one download Feb 5.

rickrams:
When a user only has queue permissions but not farm permissions, the submitter would fail when trying to get the farm display name. Now it gracefully falls back to displaying the farm ID instead.

Also fixes Houdini 21 compatibility issue where IntParmTemplate min/max parameters must be plain integers, not tuples.

Fixes: https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/336

A poor UX and error messages for users with queue only permissions

Yes

I'll need help here as i cannot run the integration tests as i've set things up.

### Is this a breaking change? No

----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/queue_parameters.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/queue_parameters.py
@@ -203,8 +203,8 @@ def _get_control_for_int_parameter(param: JobParameter) -> hou.ParmTemplate:  # 
         else param["name"]
     )
     default = _get_default_value(param)
-    min = (int(param["minValue"]),) if "minValue" in param else (0,)
-    max = (int(param["maxValue"]),) if "maxValue" in param else (10,)
+    min = int(param["minValue"]) if "minValue" in param else 0
+    max = int(param["maxValue"]) if "maxValue" in param else 10
     min_is_strict = "minValue" in param
     max_is_strict = "maxValue" in param
     hidden = _is_param_hidden(param)
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -9,6 +9,8 @@ import traceback
 from typing import Any, Dict
 from pathlib import Path
 
+from botocore.exceptions import ClientError
+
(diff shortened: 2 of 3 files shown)


leongdl (approved):

leongdl (approved):

godobyte (approved):

epmog (approved):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/queue_parameters.py
```diff
@@ -203,8 +203,8 @@ def _get_control_for_int_parameter(param: JobParameter) -> hou.ParmTemplate:  #
         else param["name"]
     )
     default = _get_default_value(param)
-    min = (int(param["minValue"]),) if "minValue" in param else (0,)
-    max = (int(param["maxValue"]),) if "maxValue" in param else (10,)
+    min = int(param["minValue"]) if "minValue" in param else 0
+    max = int(param["maxValue"]) if "maxValue" in param else 10
```
epmog, line 207: Was this actually an issue? I haven't heard of anything being wrong here. Which version was this tested with? Does it work in older versions?

rickrams: yeah, Houdini 21 kept throwing a weird error message at me until this was fixed.  Not sure if it's at all related to queue only permissions or not.

rickrams: Tested in 19.5, same error and fix worked. Change is unrelated to queue only users and just bumped into accidentally at the same time based on a specific queue config.

PR #348, opened 2026-04-21
Add job bundle to run integ tests in Deadline Cloud

jairaws:
Running the integration requires setting up Maya and its plugins in the user environment. The environment needs to have licenses configured for all the plugins (when it applies). Thus, running the full set of integration tests as part of the development cycle can be complicated.

Create a job bundle to submit a job that runs the integ tests in an real fleet. The job bundle is configured to pull the plugins from the default conda channel.

Easier to run the integ tests.

Submitted the job and verified it passed.

Yes, added README file.

### Did you modify schema files?
- [ ] Yes
- [x] No

No
----

### DIFF
--- a/job_bundle_integ_tests/README.md
+++ b/job_bundle_integ_tests/README.md
@@ -1,0 +1,52 @@
+# Houdini Integration Tests on Deadline Cloud
+
+Runs the deadline-cloud-for-houdini integration tests on a Deadline Cloud Service Managed Fleet.
+
+## Prerequisites
+
+- A Deadline Cloud farm with a queue that has the default Conda queue environment.
+- The `deadline` CLI installed and configured (`deadline config`).
+- Git LFS files resolved (expected test images are stored in LFS):
+  ```bash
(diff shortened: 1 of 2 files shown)


loachri (approved):

job_bundle_integ_tests/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Customization
+
+- To test against a different Houdini version, change the `CondaPackages` parameter (e.g., `houdini=20.5`).
+- To run only specific tests, modify the pytest marker filter in the step's run script.
+
+
+deadline bundle submit ~/workplace/jairaws/job-bundles/houdini-integ-tests-v3 \
+  --name "houdini-integ-tests" \
+  --parameter "RepoDir=/Users/ruizjair/workplace/jairaws/deadline-cloud-for-houdini" \
```
kenyrish-amazon, line 50: maybe change the personal folder path

kenyrish-amazon (approved):

PR #355, opened 2026-06-03
fix: parse sublayer nodes for path mapping properly

### LINKED ISSUE #314: Bug: sublayer node filepath not being mapped
### Describe the bug

When using a [sublayer node](https://www.sidefx.com/docs/houdini/nodes/lop/sublayer.html) with an absolute path, the filepath parameter isn't being mapped on the render workers.

### Expected Behaviour

All paths are mapped from the submitting machine to the render worker.

### Current Behaviour

In the render logs you'll get an error like below stating that the layer file can't be found and the path in the error will be the full unmapped absolute path from the submitting machine.

```
2025/10/27 09:34:47-05:00 STDERR: hou.OperationFailed: The attempted operation failed.
2025/10/27 09:34:47-05:00 STDERR: Error:       Cook error in input: /stage/karmarendersettings
2025/10/27 09:34:47-05:00 STDERR: Error:       Invalid source /stage/sublayer1.
2025/10/27 09:34:47-05:00 STDERR: (Error:       Unable to find layer file: /Users/<REDACTED>/Downloads/mug_scene/usd/assets/Mug/Mug.usd)..
2025/10/27 09:34:47-05:00 openjd_fail: Error encountered while running adaptor: Houdini exited early and did not render successfully, please check render logs. Exit code 1
```

### Reproduction Steps

1. Launch Houdini and create a new scene file
2. Go to stage
3. Add in a sublayer node and set the sublayer file path to the absolute path to some USD file. [example](https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/job_bundles/houdini_husk_usd_render/sample.usda)
5. Connect the sublayer node to a camera then to a karma render node

Image

7. In /out create a fetch node to the usd render node and feed that into the Deadline Cloud ROP

Image

9. In the deadline cloud submitter click parse files to gather all attachments and then submit the render
10. Render should fail with an error like abovedead

### Environment

At minimum:
1. Operating system: MacOS 15.7.1
2. Version of Houdini: 20.0.896
3. Version of this package: 0.7.9
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? N/A

justinsaws:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-houdini/issues/314
  
  `HOUDINI_PATHMAP` does not apply to USD/LOP node file path parameters (sublayer, reference, payload, etc.) because USD asset resolution
  uses ArResolver — not Houdini's native file I/O path mapping. This means when a scene with LOP sublayer or reference nodes using absolute
  paths is submitted to Deadline Cloud, the worker receives the submitting machine's local paths unchanged, causing render failure with
  `Unable to find layer file`.
  
  Added `_remap_lop_file_paths()` in `houdini_handler.py` that runs after scene load (alongside the existing `_path_map_envs()`). It iterates
  all LOP nodes in the scene via `hou.node("/").recursiveGlob("*", filter=hou.nodeTypeFilter.Lop)` and applies `pathmap -t` to any string
  parameter that is either:
  - Tagged as `hou.stringParmType.FileReference` (e.g., sublayer's `filepath1`), or
  - Named `filepath*` (e.g., reference's `filepath1`, which SideFX inconsistently typed as `stringParmType.Regular`)
  
  Expression-based paths (`$HIP/...`, backtick expressions) are skipped to avoid clobbering variable references. Single quotes in paths are
  escaped to prevent hscript command injection.
  
  LOP sublayer, reference, and other file-path-bearing LOP nodes will now have their paths correctly remapped on Deadline Cloud workers when
  `HOUDINI_PATHMAP` is configured via path mapping rules. No impact on non-LOP workflows — the method early-returns if `HOUDINI_PATHMAP` is
  not set.
  
  **Known limitation**: Absolute paths *inside* referenced USD files (e.g., composition arcs within a `.usd` file on disk) are not addressed
  by this fix — those would require a custom USD ArResolver plugin.
  
  - Unit tests: 9 new tests in `TestRemapLopFilePaths` covering early return, successful remap (FileReference and filepath-named parms),
  empty values, expression skipping, non-string parm skipping, no-op on unchanged paths, and hscript error handling.
  - Integration tests: Submitted test scenes to Deadline Cloud using adaptor wheels with the fix. Verified on Houdini 19.5, 20.0, and 20.5 that
  sublayer/reference paths are remapped and renders complete successfully.
  
  All 167 unit tests pass. Integration test confirmed render success on farm with path mapping rule
  `/Users/justins/dev/github/aws-deadline/deadline-cloud-for-houdini` → `/sessions/.../assetroot-...`. Adaptor output shows:
  
  Remapped LOP parm /stage/sublayer_mug/filepath1: /Users/justins/.../Mug.usda -> /sessions/.../Mug.usda
  Remapped LOP parm /stage/sublayer_chair/filepath1: /Users/justins/.../Chair.usda -> /sessions/.../Chair.usda
  
  N/A — no files added/removed, no installer changes.
  
  - Docstring added to `_remap_lop_file_paths()` explaining why it exists (HOUDINI_PATHMAP/ArResolver gap) and what it does.
  - No README/schema changes needed — this is an internal adaptor behavior fix with no interface changes.
  
  No. This is an additive fix that runs transparently during scene load. No changes to init-data, run-data, or the adaptor interface. Jobs
  submitted with older versions of the submitter will benefit from this fix on workers running the updated adaptor.
  
  ----
  
  *By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your
  choice.*

### DIFF
--- a/src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
+++ b/src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
@@ -11,7 +11,7 @@ except ImportError:  # pragma: no cover
     raise OSError("Could not find the Houdini module. Are you running this inside of Houdini?")
 
 if TYPE_CHECKING:  # pragma: no cover
-    from hou import Node
+    from hou import Node, Parm, ParmTemplate
 
 
 class HoudiniHandler:
@@ -65,6 +65,39 @@ class HoudiniHandler:
 
         # Reload the env variables
         hou.hscript("varchange")
+
+    def _remap_lop_file_paths(self) -> None:
+        """
+        Applies HOUDINI_PATHMAP to LOP node file path parameters.
+
+        HOUDINI_PATHMAP does not apply to USD/LOP nodes (sublayer, reference, etc.)
+        because USD asset resolution uses ArResolver, not Houdini's file I/O layer.
+        This method manually remaps any file reference string parms on LOP nodes.
+        """
+        if not os.environ.get("HOUDINI_PATHMAP"):
+            return
+
+        lop_nodes: list[Node] = hou.node("/").recursiveGlob("*", filter=hou.nodeTypeFilter.Lop)
+        for node in lop_nodes:
+            parms: list[Parm] = node.parms()
+            for parm in parms:
+                template: ParmTemplate = parm.parmTemplate()
+                if template.type() != hou.parmTemplateType.String:
+                    continue
+                if (
+                    template.stringType() == hou.stringParmType.FileReference
+                    or parm.name().startswith("filepath")
+                ):
+                    original: str = parm.unexpandedString()
+                    if not original or original.startswith("$") or original.startswith("`"):
+                        continue
+                    mapped, err = hou.hscript(f"pathmap -t '{original}' -c")
+                    mapped = mapped.strip()
+                    if err:
+                        print(f"Error remapping {parm.path()}: {err}")
+                    elif mapped and mapped != original:
+                        parm.set(mapped)
+                        print(f"Remapped LOP parm {parm.path()}: " f"{original} -> {mapped}")
 
     def set_node_settings(self, node):
         # this is a place holder function
@@ -236,3 +269,4 @@ class HoudiniHandler:
             print(e)
 
         self._path_map_envs()
+        self._remap_lop_file_paths()
--- a/test/unit/deadline_adaptor_for_houdini/unit/HoudiniClient/mock_hou.py
+++ b/test/unit/deadline_adaptor_for_houdini/unit/HoudiniClient/mock_hou.py
@@ -20,3 +20,6 @@ setattr(this_module, "hscript", Mock(name=module_name + ".hscript"))
 setattr(
     this_module, "applicationVersionString", Mock(name=module_name + ".applicationVersionString")
 )
+setattr(this_module, "nodeTypeFilter", Mock(name=module_name + ".nodeTypeFilter"))
+setattr(this_module, "parmTemplateType", Mock(name=module_name + ".parmTemplateType"))
+setattr(this_module, "stringParmType", Mock(name=module_name + ".stringParmType"))
--- a/test/unit/deadline_adaptor_for_houdini/unit/HoudiniClient/test_handler.py
+++ b/test/unit/deadline_adaptor_for_houdini/unit/HoudiniClient/test_handler.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from unittest.mock import Mock, patch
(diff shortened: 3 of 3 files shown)


src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
```diff
@@ -66,6 +66,38 @@ def _path_map_envs(self) -> None:
...
+        HOUDINI_PATHMAP does not apply to USD/LOP nodes (sublayer, reference, etc.)
+        because USD asset resolution uses ArResolver, not Houdini's file I/O layer.
+        This method manually remaps any file reference string parms on LOP nodes.
+        """
+        if not os.environ.get("HOUDINI_PATHMAP"):
+            return
+
+        for node in hou.node("/").recursiveGlob("*", filter=hou.nodeTypeFilter.Lop):
+            for parm in node.parms():
+                template = parm.parmTemplate()
```
jairaws: nit: What kind of object is `template`? At first I thought it was a string, but looking at the logic below, it seems to be a Houdini template type? IMO, the code would be easier to understand if we typed these variables.

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
```diff
@@ -66,6 +66,38 @@ def _path_map_envs(self) -> None:
...
+                if template.type() != hou.parmTemplateType.String:
+                    continue
+                if (
+                    template.stringType() == hou.stringParmType.FileReference
+                    or parm.name().startswith("filepath")
+                ):
+                    original = parm.unexpandedString()
+                    if not original or original.startswith("$") or original.startswith("`"):
+                        continue
+                    escaped = original.replace("'", "'\"'\"'")
```
jairaws: Q: It is not clear to me why we are replacing single quotes with `'\"'\"'`. If I am reading correctly, we are mapping `'` to `'"'"'`?  Could explain what's the purpose of this mapping?

I also wonder if this will work in every OS.

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
justinsaws: Fixed.

src/deadline/houdini_adaptor/HoudiniClient/houdini_handler.py
justinsaws: This was escaping single quotes in case they appeared in the path but I've now changed it to match the same path mapping mechanisms we use in other places in this file.

jairaws (approved):

ttblanchard (approved):

PR #360, opened 2026-07-02
feat: add headless HoudiniSubmitter on the unified BaseSubmitter

leon-li-inspire:
Fixes: N/A (no tracking issue; enables the AYON Houdini integration on the unified submitter API)

Host-agnostic consumers — notably the AYON bridge, which resolves submitters via `get_submitter_for_host("houdini")` — need to drive Houdini submission headlessly, with no GUI. That requires a Houdini implementation of the unified `BaseSubmitter` ABC (`deadline.client.api`, aws-deadline/deadline-cloud#1245): `get_settings` / `get_job_template` / `get_parameter_values` / `get_asset_references`.

The Houdini submitter had no such engine. It was also just re-migrated to the shared Qt dialog in #372, so a naive second implementation would duplicate the node-driven bundle builders and risk the GUI and headless paths drifting apart.

Add a headless `HoudiniSubmitter(BaseSubmitter)` + `HoudiniSubmitterSettings(BaseSubmitterSettings)` in `submitter.py`, layered on top of #372 as a thin adapter over its existing node-driven builders, and route the GUI's own bundle path through the same engine — so both paths share one implementation and emit identical bundles.

- **`deadline_cloud_for_houdini/submitter.py`**
  - New `HoudiniSubmitter(BaseSubmitter)` + `HoudiniSubmitterSettings(BaseSubmitterSettings)`. The submitter resolves everything from a single ROP node path (`set_rop_node_path()` / constructor arg).
  - `get_settings` scans the scene for the real input/output asset references (same detection as "Parse Files") so consumers reading `settings.*` (e.g. AYON's create parameterDefinitions UI and its publish-time job-attachments collection) get populated directories rather than empty defaults. It takes a `scan_assets: bool = True` flag; the GUI submit/export path passes `False` to skip the scene/USD walk it would only discard.
  - `get_job_template` delegates to the node-driven `_get_job_template` for the ROP-derived step graph and overrides the settings-driven scalars (job name/description); it also reconciles the `AdaptorWheels` parameterDefinition + `OverrideAdaptor` jobEnvironment with the settings-driven value so the template and parameter values stay consistent.
  - `get_parameter_values` sources scalar `deadline:*` values from settings and merges caller-supplied `queue_parameters` with **caller-wins** precedence, de-duping the ROP's own values. An empty/`None` caller value is treated as "not provided", so a bare queue default never blanks a populated node parameter (e.g. the computed `CondaPackages`).
  - `get_asset_references` returns the typed `AssetReferences` per the base contract.
  - `_create_job_bundle` (the GUI submit/export callback path from #372) now builds the template + parameter values by driving `HoudiniSubmitter` instead of calling the module builders directly. The dialog-provided `asset_references` are still written as-is (they carry the user's Job Attachments edits).
- **`pyproject.toml`**: pin `deadline >= 0.60.2, < 0.61` — `0.60.2` is the first release carrying `BaseSubmitter` (`0.60.1` has none).

- AYON (and any other host-agnostic consumer) can drive Houdini submission headlessly through the unified `BaseSubmitter` API.
- The native Houdini GUI submit/export path and the headless path now share a single bundle-building implementation, so there is one source of truth for bundle contents (no GUI/headless drift).
- Additive only — no change to the #372 GUI behavior; the GUI bundle output is byte-identical before/after routing through the engine.

**Unit** — Full submitter unit suite passes against `deadline 0.60.2`, including the new engine tests in `test/unit/.../test_houdini_submitter.py` (ABC conformance, `set_rop_node_path`, `get_settings` frame-range + scene-ref population, `scan_assets=False` skip, builder delegation, settings-driven overrides, queue-param caller-wins merge incl. the empty-caller-value regression, adaptor-wheels consistency, missing-ROP error). `ruff` / `black` / `mypy` clean. Verified the GUI bundle output is byte-identical before/after routing through `HoudiniSubmitter`.

Ran the repo's submitter integration suite (`test/integ`, `-m submitter`) against Houdini **21.0.631** (hython), per `.kiro/powers/houdini-dev-power/steering/integration-testing.md`. These tests launch hython with each test scene, build a job bundle via the routed `HoudiniSubmitter` path, validate the template with `openjd check`, and compare against the expected bundle. Run logged out of Deadline Cloud (isolated empty config) so live queue parameters don't leak into the generated templates.

```
test/integ/test_houdini_submitters.py::TestSubmitters::test_minimal_scene_submitter        PASSED
test/integ/test_houdini_submitters.py::TestSubmitters::test_wedge_node_submitter            PASSED
test/integ/test_houdini_submitters.py::TestSubmitters::test_render_dependencies_submitter   PASSED
test/integ/test_houdini_submitters.py::TestSubmitters::test_usd_scene_dependency_detection  PASSED
============================== 4 passed in 9.02s ==============================
```

N/A — `installer/` was not modified, and no file was added or removed from `src/` (only `src/.../submitter.py` was modified in place; the new file is a unit test under `test/`).

Yes — docstrings were added/updated for `HoudiniSubmitter`, `HoudiniSubmitterSettings`, and all four base-contract methods (including the `scan_assets` flag on `get_settings`). No README/DEVELOPMENT/adaptor-schema changes are needed: this change is submitter-only and does not touch the adaptor's init-data/run-data.

No. The change is additive — a new headless engine plus internal routing of the existing GUI bundle path through it. No public adaptor init-data/run-data contract is changed, and the GUI submit/export output is byte-identical. The only dependency change is raising the `deadline` floor to `>= 0.60.2` (first release carrying `BaseSubmitter`).

## Depends on
aws-deadline/deadline-cloud#1245 (the `BaseSubmitter` / `BaseSubmitterSettings` base contract, re-exported from `deadline.client.api`), released in `deadline` `0.60.2`.

----

### DIFF
--- a/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
+++ b/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
@@ -1,5 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
+from dataclasses import dataclass
 from enum import Enum
 import copy
 import os
@@ -7,11 +8,12 @@ import sys
 import yaml
 import json
 import traceback
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, cast
 from pathlib import Path
 
 from botocore.exceptions import ClientError
 
+from deadline.client.api import BaseSubmitter, BaseSubmitterSettings
 from deadline.client.job_bundle._yaml import deadline_yaml_dump
 from deadline.client import api
 from deadline.client.job_bundle.submission import AssetReferences
@@ -51,6 +53,258 @@ _REFRESHING_TEXT = "<refreshing>"
 class RenderStrategy(Enum):
     SEQUENTIAL = "SEQUENTIAL"
     PARALLEL = "PARALLEL"
+
+
+@dataclass
+class HoudiniSubmitterSettings(BaseSubmitterSettings):
+    """Houdini-specific submission settings for the unified :class:`BaseSubmitter`.
+
+    The Houdini submitter derives the job's step graph from the live ROP node
+    (via the module-level ``_get_job_template``). Beyond the DCC-agnostic base
+    fields (``job_name``, ``description``, ``priority``, ``initial_status``,
+    ``max_failed_tasks_count``, ``max_retries_per_task``, the asset-reference
+    lists) this carries the ROP node path plus the Houdini-specific
+    adaptor-wheels fields, so the build methods can honor consumer edits to
+    those values rather than silently re-reading the node.
+
+    Note: the *frame range* is intentionally not a settings-driven input. A
+    Houdini job's per-step frames/dependencies are derived from the live ROP
+    network via ``hscript render`` (see ``_get_rop_steps``); a single
+    ``frame_list`` string cannot represent that graph, so ``get_settings``
+    populates ``frame_list`` for information only and the builders take frames
+    from the ROP.
+    """
+
+    rop_node_path: str = ""
+    include_adaptor_wheels: bool = False
+    adaptor_wheels: str = ""
+
+
+class HoudiniSubmitter(BaseSubmitter):
+    """Headless :class:`BaseSubmitter` implementation for Houdini.
+
+    This is the unified-API entry point that host-agnostic consumers (e.g. the
+    AYON bridge, ``get_submitter_for_host("houdini")``) drive, and the same
+    engine the native GUI submitter routes through (see ``_create_job_bundle``).
+    ``get_job_template`` reuses the module-level ``_get_job_template`` for the
+    ROP-derived step graph and overrides the settings-driven fields; parameter
+    values are built from ``settings`` plus the ROP's queue parameters, and
+    asset references come from the ``_assets`` scene scan. The submitter
+    resolves everything from a single ROP node path; call
+    :meth:`set_rop_node_path` (or construct with one) before requesting
+    settings/templates.
+    """
+
+    def __init__(self, rop_node_path: str = "") -> None:
+        self._rop_node_path = rop_node_path
+
+    def set_rop_node_path(self, rop_node_path: str) -> None:
+        """Point the submitter at the Deadline Cloud ROP node to submit.
+
+        Lets external callers (e.g. the AYON create/publish plugins) seed the
+        node without reaching into the private attribute.
+        """
+        self._rop_node_path = rop_node_path
+
+    def _get_rop_node(self) -> Optional["hou.Node"]:
+        if self._rop_node_path:
+            return hou.node(self._rop_node_path)
+        return None
+
+    def _require_rop(self) -> "hou.Node":
+        rop = self._get_rop_node()
+        if rop is None:
+            raise RuntimeError(f"Cannot find ROP node at path: {self._rop_node_path}")
+        return rop
+
+    def get_settings(self, scan_assets: bool = True) -> HoudiniSubmitterSettings:
+        """Collect submission settings from the live scene / ROP node.
+
+        Populates the base-contract fields (job name, frame range, and the
+        input/output asset references discovered by walking the scene) so that
+        consumers reading them off the returned settings — e.g. the AYON create
+        plugin's parameterDefinitions UI and its publish-time job-attachments
+        collection — see real values rather than empty defaults. The scene scan
+        is the same one the native "Parse Files" button uses.
+
+        ``scan_assets`` (default ``True``, so base-contract callers are
+        unchanged) gates the input/output asset-reference scan. Callers that do
+        not read ``settings.input_*``/``output_directories`` — notably the GUI
+        submit/export path, which writes ``asset_references.yaml`` from the
+        dialog-provided references — can pass ``False`` to skip a potentially
+        expensive filesystem + USD walk whose output would be discarded.
+        """
+        settings = HoudiniSubmitterSettings()
+        hip_file = _get_hip_file()
+        settings.job_name = hou.hipFile.basename() or "Untitled"
+        settings.project_path = hou.getenv("HIP", "") or ""
+        # Baseline inputs: the hip file only. Overwritten below by the scene scan
+        # when it runs and succeeds; otherwise this fallback stands (no ROP, scan
+        # skipped, or scan failed).
+        settings.input_filenames = [hip_file] if hip_file else []
+
+        # frame_list from the playbar; refined from the ROP's frame-range tuple
+        # below when a node is available.
+        playbar = hou.playbar.frameRange()
+        settings.frame_list = f"{int(playbar[0])}-{int(playbar[1])}"
+        settings.output_path = settings.project_path
+
+        rop = self._get_rop_node()
+        if rop is not None:
+            settings.rop_node_path = rop.path()
+            if rop.parm("name"):
+                settings.job_name = rop.parm("name").evalAsString() or settings.job_name
+            if rop.parm("priority"):
+                settings.priority = rop.parm("priority").eval()
+            if rop.parm("initial_status"):
+                settings.initial_status = rop.parm("initial_status").evalAsString()
+            if rop.parm("failed_tasks_limit"):
+                settings.max_failed_tasks_count = rop.parm("failed_tasks_limit").eval()
+            if rop.parm("task_retry_limit"):
+                settings.max_retries_per_task = rop.parm("task_retry_limit").eval()
+            if rop.parm("description"):
+                settings.description = rop.parm("description").evalAsString()
+            if rop.parm("include_adaptor_wheels"):
+                settings.include_adaptor_wheels = bool(rop.parm("include_adaptor_wheels").eval())
+            if rop.parm("adaptor_wheels"):
+                settings.adaptor_wheels = rop.parm("adaptor_wheels").evalAsString()
+
+            frame_range = rop.parmTuple("f")
+            if frame_range:
+                start = int(frame_range[0].eval())
+                end = int(frame_range[1].eval())
+                step = int(frame_range[2].eval()) if len(frame_range) > 2 else 1
+                settings.frame_list = f"{start}-{end}:{step}" if step != 1 else f"{start}-{end}"
+
+            # Discover the real input/output asset references by scanning the
+            # scene (same detection as "Parse Files"), so consumers that read
+            # these off the settings get populated directories even when the
+            # ROP's Job Attachments multiparms have not been parsed yet. Skipped
+            # when scan_assets is False (the caller does not read these fields);
+            # on failure the hip-file baseline set above stands.
+            if scan_assets:
+                try:
+                    refs = _get_scene_asset_references(rop)
+                    settings.input_filenames = sorted(refs.input_filenames)
+                    settings.input_directories = sorted(refs.input_directories)
+                    settings.output_directories = sorted(refs.output_directories)
+                    if settings.output_directories:
+                        settings.output_path = settings.output_directories[0]
+                except Exception:
+                    # Scene scan can fail (e.g. missing/locked nodes); keep the
+                    # hip-file baseline rather than breaking settings collection.
+                    pass
+
+        return settings
+
+    def get_job_template(
+        self,
+        settings: BaseSubmitterSettings,
+        host_requirements: Optional[dict[str, Any]] = None,
+    ) -> dict[str, Any]:
+        houdini_settings = cast(HoudiniSubmitterSettings, settings)
+        rop = self._require_rop()
+
+        # The step graph (per-step frames, dependencies, wedges) is derived from
+        # the live ROP network by _get_job_template and cannot be reconstructed
+        # from settings; the job name/description are settings-driven so a
+        # consumer's edits are honored (matching deadline-cloud-for-maya).
+        job_template = _get_job_template(rop, host_requirements)
+
+        if houdini_settings.job_name:
+            job_template["name"] = houdini_settings.job_name
+        if houdini_settings.description:
+            job_template["description"] = houdini_settings.description
+        elif "description" in job_template and not houdini_settings.description:
+            # An explicitly-cleared description should not fall back to the node.
+            del job_template["description"]
+
+        # Reconcile the adaptor-wheels override with settings so the template's
+        # AdaptorWheels parameterDefinition/jobEnvironment stays consistent with
+        # the AdaptorWheels *value* get_parameter_values emits (which is
+        # settings-driven). _get_job_template added the override based on the
+        # node; a headless consumer that edited the wheels settings could
+        # otherwise desync (definition without value, or value without
+        # definition -> CreateJob reject).
+        wheels_enabled = houdini_settings.include_adaptor_wheels and os.path.exists(
+            houdini_settings.adaptor_wheels
+        )
+        _apply_adaptor_wheels_override(job_template, wheels_enabled)
+
+        return job_template
+
+    def get_parameter_values(
+        self,
+        settings: BaseSubmitterSettings,
+        queue_parameters: list[dict[str, Any]],
+    ) -> list[dict[str, Any]]:
+        houdini_settings = cast(HoudiniSubmitterSettings, settings)
+        rop = self._require_rop()
+
+        # Scalar job parameters come from settings (honoring consumer edits); the
+        # ROP is still consulted for its per-node queue parameter values and hip
+        # file, which have no settings equivalent.
+        parameter_values: list[dict[str, Any]] = [
+            {"name": "deadline:priority", "value": houdini_settings.priority},
+            {"name": "deadline:targetTaskRunStatus", "value": houdini_settings.initial_status},
+            {
+                "name": "deadline:maxFailedTasksCount",
+                "value": houdini_settings.max_failed_tasks_count,
+            },
+            {"name": "deadline:maxRetriesPerTask", "value": houdini_settings.max_retries_per_task},
+            {"name": "HipFile", "value": _get_hip_file()},
+            *get_queue_parameter_values_as_openjd(rop),
+        ]
+        if houdini_settings.include_adaptor_wheels and os.path.exists(
+            houdini_settings.adaptor_wheels
+        ):
+            parameter_values.append(
+                {"name": "AdaptorWheels", "value": houdini_settings.adaptor_wheels}
+            )
+
+        # Merge caller-supplied queue_parameters with caller-wins precedence.
+        # The node already emits its own queue parameters (CondaPackages,
+        # CondaChannels, ...) via get_queue_parameter_values_as_openjd, and a
+        # headless consumer (e.g. AYON) passes the SAME queue parameters here as
+        # its resolved/edited values -- so the two sets overlap by design. The
+        # caller's value supersedes the node's on a name collision (mirroring how
+        # the GUI's set_queue_parameter_values_from_openjd writes dialog edits
+        # back onto the node before the bundle is read). A plain de-dup also
+        # avoids the duplicate parameterValues that CreateJob rejects.
+        #
+        # An empty/None caller value is treated as "not provided": consumers such
+        # as AYON build this list from get_queue_parameters(), which resolves
+        # every parameter to its queue default -- and a queue whose CondaPackages
+        # default is empty would otherwise clobber the node's computed value
+        # (e.g. "houdini=20.5.* houdini-openjd=0.x.*"). We only let a caller value
+        # win when it actually carries a value, so a bare queue default never
+        # blanks a populated node parameter. Falsy-but-real values (0, False) are
+        # preserved.
+        caller_values = {
+            param["name"]: param["value"]
+            for param in queue_parameters
+            if "value" in param and param["value"] not in (None, "")
+        }
+        merged: list[dict[str, Any]] = [
+            {"name": param["name"], "value": caller_values.get(param["name"], param["value"])}
+            for param in parameter_values
+        ]
+        existing_names = {param["name"] for param in merged}
+        merged.extend(
+            {"name": name, "value": value}
+            for name, value in caller_values.items()
+            if name not in existing_names
+        )
+        return merged
+
+    def get_asset_references(self, settings: BaseSubmitterSettings) -> AssetReferences:
+        rop = self._require_rop()
+        # Scan the scene for the authoritative asset references rather than
+        # reading the ROP's (possibly unparsed) Job Attachments multiparms, so a
+        # headless consumer gets real input/output paths without a prior
+        # "Parse Files" pass. Returns the typed AssetReferences per the
+        # BaseSubmitter contract; callers serialize with .to_dict().
+        return _get_scene_asset_references(rop)
 
 
 def _get_houdini_version() -> str:
@@ -266,34 +520,12 @@ def _get_render_strategy_for_node(node: hou.Node) -> RenderStrategy:
     return render_strategy
 
 
-def _get_parameter_values(node: hou.Node) -> dict[str, Any]:
-    priority = node.parm("priority").eval()
-    initial_status = node.parm("initial_status").evalAsString()
-    failed_tasks_limit = node.parm("failed_tasks_limit").eval()
-    task_retry_limit = node.parm("task_retry_limit").eval()
-    parameter_values = [
-        {"name": "deadline:priority", "value": priority},
-        {"name": "deadline:targetTaskRunStatus", "value": initial_status},
-        {"name": "deadline:maxFailedTasksCount", "value": failed_tasks_limit},
-        {"name": "deadline:maxRetriesPerTask", "value": task_retry_limit},
-        {"name": "HipFile", "value": _get_hip_file()},
-        *get_queue_parameter_values_as_openjd(node),
-    ]
-
-    if node.parm("include_adaptor_wheels").eval():
-        parameter_values.append(
-            {"name": "AdaptorWheels", "value": node.parm("adaptor_wheels").evalAsString()}
-        )
-
-    return {"parameterValues": parameter_values}
-
-
 def read_ui_settings_from_node(node: hou.Node) -> HoudiniSubmitterUISettings:
     """Populate a HoudiniSubmitterUISettings from the Deadline Cloud ROP node's parameters.
 
     This lets the shared SubmitJobToDeadlineDialog (used by the Python panel) open with the
     node's configured values instead of pure dataclass defaults. Parameter names match those
-    defined in the HDA DialogScript and read elsewhere in this module (e.g. _get_parameter_values).
+    defined in the HDA DialogScript and read elsewhere in this module (e.g. get_settings).
 
     Note the deliberate name remapping: the node exposes ``failed_tasks_limit`` /
     ``task_retry_limit``, but deadline-cloud's SharedJobPropertiesWidget reads
@@ -474,6 +706,51 @@ def _get_job_template(
     return job_template
 
 
+def _load_adaptor_override_environment() -> dict[str, Any]:
+    override_file = os.path.join(os.path.dirname(__file__), "adaptor_override_environment.yaml")
+    with open(override_file) as yaml_file:
+        return yaml.safe_load(yaml_file)
+
+
+# Names contributed by adaptor_override_environment.yaml, used to detect/strip
+# the override so the template can be reconciled with the settings-driven state.
+_ADAPTOR_OVERRIDE_PARAM_NAMES = {"AdaptorWheels", "OverrideAdaptorName"}
+_ADAPTOR_OVERRIDE_ENV_NAME = "OverrideAdaptor"
+
+
+def _apply_adaptor_wheels_override(job_template: dict[str, Any], enabled: bool) -> None:
+    """Add or remove the adaptor-wheels override on ``job_template`` in place.
+
+    Idempotent: makes the AdaptorWheels ``parameterDefinition``s and the
+    ``OverrideAdaptor`` ``jobEnvironment`` present when ``enabled`` and absent
+    otherwise, regardless of what the node-driven ``_get_job_template`` already
+    added. This keeps the template consistent with the settings-driven
+    AdaptorWheels parameter *value* emitted by ``get_parameter_values``.
+    """
+    already_present = any(
+        pd.get("name") in _ADAPTOR_OVERRIDE_PARAM_NAMES
+        for pd in job_template.get("parameterDefinitions", [])
+    )
+
+    if enabled and not already_present:
+        override = _load_adaptor_override_environment()
+        job_template.setdefault("parameterDefinitions", []).extend(override["parameterDefinitions"])
+        job_template.setdefault("jobEnvironments", []).append(override["environment"])
+    elif not enabled and already_present:
+        job_template["parameterDefinitions"] = [
+            pd
+            for pd in job_template.get("parameterDefinitions", [])
+            if pd.get("name") not in _ADAPTOR_OVERRIDE_PARAM_NAMES
+        ]
+        job_template["jobEnvironments"] = [
+            env
+            for env in job_template.get("jobEnvironments", [])
+            if env.get("name") != _ADAPTOR_OVERRIDE_ENV_NAME
+        ]
+        if not job_template["jobEnvironments"]:
+            del job_template["jobEnvironments"]
+
+
 def _get_step_template(node: Dict, ignore_input_nodes: bool):
     init_data = {
         "scene_file": "{{Param.HipFile}}",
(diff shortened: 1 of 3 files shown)


src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter_api.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+            settings.separate_steps = bool(
+                rop.parm("separate_steps").eval() if rop.parm("separate_steps") else False
+            )
+
+            frame_range = rop.parmTuple("f")
+            if frame_range:
+                start = int(frame_range[0].eval())
+                end = int(frame_range[1].eval())
+                settings.frame_list = f"{start}-{end}"
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in b527f4b — the ROP frame-range step is now preserved in OpenJD frame-list syntax (`start-end:step`) when the increment isn't 1. Covered by `test_frame_list_includes_step_when_non_one`.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter_api.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+            settings.max_failed_tasks_count = (
+                rop.parm("failed_tasks_limit").eval() if rop.parm("failed_tasks_limit") else 20
+            )
+            settings.max_retries_per_task = (
+                rop.parm("task_retry_limit").eval() if rop.parm("task_retry_limit") else 5
+            )
+            settings.separate_steps = bool(
+                rop.parm("separate_steps").eval() if rop.parm("separate_steps") else False
+            )
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in b527f4b — `get_settings` now populates `include_adaptor_wheels` and `description` from the ROP parms instead of leaving them at dataclass defaults. Covered by `test_node_backed_fields_populated`.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter_api.py
```diff
@@ -0,0 +1,164 @@
...
+                # Output resolution can fail (e.g. missing/locked nodes); fall back
+                # to the hip directory already assigned above rather than breaking
+                # settings collection.
+                pass
+
+        return settings
+
+    def get_job_template(
+        self,
+        settings: SubmitterSettings,
```
crowecawcaw: It looks like settings are ignored and the template is generated only from the scene. Other submitters use the settings from get_settings(). Should this be organized differently?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -849,3 +486,174 @@ def get_houdini_environments(init_data_attachment: dict[str, Any]) -> list[dict[
             },
         }
     ]
+
+
+class HoudiniSubmitter(BaseSubmitter):
+    """BaseSubmitter implementation for Houdini submissions."""
+
+    def __init__(self, rop_node_path: str = ""):
+        self._rop_node_path = rop_node_path
```
leon-li-inspire: Please provide one method can be used by external caller to set `_rop_node_path`.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -849,3 +486,174 @@ def get_houdini_environments(init_data_attachment: dict[str, Any]) -> list[dict[
...
+            # for the scalar output_path so the ABC field reflects a real path.
+            try:
+                asset_refs = _get_evaluated_asset_references(rop)
+                output_directories = sorted(asset_refs.output_directories)
+                if output_directories:
+                    settings.output_directories = output_directories
+                    settings.output_path = output_directories[0]
+            except Exception:
+                # Output resolution can fail (e.g. missing/locked nodes); fall back
+                # to the hip directory already assigned above rather than breaking
```
leon-li-inspire: Don't we need to have a log here ?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -252,92 +266,38 @@ def _get_render_strategy_for_node(node: hou.Node) -> RenderStrategy:
     return render_strategy
 
 
-def _get_parameter_values(node: hou.Node) -> dict[str, Any]:
-    priority = node.parm("priority").eval()
-    initial_status = node.parm("initial_status").evalAsString()
-    failed_tasks_limit = node.parm("failed_tasks_limit").eval()
-    task_retry_limit = node.parm("task_retry_limit").eval()
+def _get_parameter_values(rop: hou.Node, settings: HoudiniSubmitterSettings) -> dict[str, Any]:
```
leon-li-inspire: Is this duplicate with the `get_parameter_values` in `HoudiniSubmitter` ?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -252,92 +266,38 @@ def _get_render_strategy_for_node(node: hou.Node) -> RenderStrategy:
...
-        )
+    if settings.include_adaptor_wheels:
+        parameter_values.append({"name": "AdaptorWheels", "value": settings.adaptor_wheels})
 
     return {"parameterValues": parameter_values}
 
 
-def _follow_fetch_nodes(node: hou.Node) -> hou.Node:
-    """Follow a chain of fetch nodes to find the final target node.
+def _get_job_template(rop: hou.Node, settings: HoudiniSubmitterSettings) -> dict[str, Any]:
```
leon-li-inspire: Is it duplicate with the `get_job_template` method in `HoudiniSubmitter`?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -849,3 +486,174 @@ def get_houdini_environments(init_data_attachment: dict[str, Any]) -> list[dict[
...
+        asset_references = _get_evaluated_asset_references(rop)
+        return asset_references.to_dict()
+
+    def _get_rop_node(self):
+        if self._rop_node_path:
+            return hou.node(self._rop_node_path)
+        return None
+
+
+def _create_job_bundle(
```
leon-li-inspire: This method should be moved back to `gui_submitter.py` as it is the place this method will be used ?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Done in 9d0652d — added `HoudiniSubmitter.set_rop_node_path(rop_node_path)` so external callers set the node path through a public method instead of touching `_rop_node_path`. Covered by `test_set_rop_node_path_updates_resolved_node`.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Done in 9d0652d — the except now emits `logger.warning(..., exc_info=True)` (module logger added via `getLogger(__name__)`) noting the ROP whose output resolution failed and the fallback to the hip directory, instead of a bare `pass`.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Not a duplicate — `get_parameter_values` is the public `BaseSubmitter` override (resolves the ROP, merges caller-supplied `queue_parameters`, de-dups), and `_get_parameter_values` was the pure settings->parameterValues builder it delegated to. To make that relationship obvious I moved it onto the class as a `@staticmethod` in 9d0652d (matching deadline-cloud-for-maya / -nuke), so the builder and its wrapper now read as one unit.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Same as the `_get_parameter_values` thread: not a duplicate — the method is the public override (adds host requirements) and `_get_job_template` is the pure builder. Folded onto the class as a `@staticmethod` in 9d0652d to match Maya/Nuke and remove the free-function-that-looks-like-a-duplicate.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: I'd keep it in the engine (`submitter.py`). The four hython integration scripts (`test/integ/test_scripts/*/_test_hip.py`) import `_create_job_bundle` from `deadline_cloud_for_houdini.submitter` and run under headless `hython` — importing it from `gui_submitter` would drag in `SubmitJobProgressDialog`/`qtpy` (verified: `gui_submitter` raises `QtBindingsNotFoundError` without Qt bindings), breaking those tests. It also builds the bundle purely by driving `HoudiniSubmitter`, so it's engine logic with no GUI dependency. `gui_submitter`'s callbacks import and call it. Happy to revisit if you'd prefer a third headless module for it.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter_api.py
leon-li-inspire: Fixed — the builders are now settings-driven. `get_job_template`/`get_parameter_values` read job name, description, priority, statuses, retries, separate_steps and adaptor-wheels off `HoudiniSubmitterSettings` (staticmethods `_get_job_template`/`_get_parameter_values` on the class), instead of re-deriving everything from the ROP. The ROP is still walked for its step network, per-node queue-parameter values and hip file. This matches how the other submitters consume `get_settings()` output.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: You're right — I've now removed the redundancy in edf52f9. The `_get_parameter_values` staticmethod is gone; its body is inlined into the public `get_parameter_values`. (My earlier staticmethod split copied Maya, but Maya's staticmethods consume precomputed shared render-layer data, whereas here the builder was only ever called by its own wrapper with the same inputs — so it was pure indirection.)

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Same here — removed in edf52f9. `_get_job_template` is inlined into the public `get_job_template` (which still resolves the ROP once and applies host_requirements). No more staticmethod/instance-method pair.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -849,3 +409,259 @@ def get_houdini_environments(init_data_attachment: dict[str, Any]) -> list[dict[
...
+    rop_node: hou.Node, job_bundle_dir: str, asset_references: AssetReferences
```
leon-li-inspire: Do we really need asset_references as the input here, couldn't we get this directly from `submitter.get_asset_references` ?

leon-li-inspire: Done in 4f5dc19 — removed the `asset_references` parameter. `_create_job_bundle` now gets all three pieces from the one submitter: `get_job_template` / `get_parameter_values` / `get_asset_references`. Every caller had been passing exactly `_get_evaluated_asset_references(node)`, so the argument was redundant. As a bonus it fixes a latent staleness bug in `submit_callback`, which computed `asset_references` *before* a possible `_parse_files()` pass and passed the pre-parse value — the written `asset_references.yaml` now reflects the node's state at bundle time. GUI callbacks and the four hython integ scripts updated; `submit_callback` keeps its own local only for the hip-in-references check.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -1,41 +1,60 @@
...
-from deadline.client.job_bundle import create_job_history_bundle_dir
-from deadline.client.job_bundle.parameters import JobParameter
-from deadline.client.config import get_setting
-from deadline.client.config.config_file import str2bool
-from deadline.job_attachments.upload import S3AssetManager
-from deadline.job_attachments.models import JobAttachmentS3Settings
 
-from .queue_parameters import update_queue_parameters, get_queue_parameter_values_as_openjd
-from ._assets import _get_hip_file, _get_evaluated_asset_references, _parse_files
+import hou  # type: ignore[import]
```
leon-li-inspire: Why do we need to have `#type: ignore[import]` here ?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/gui_submitter.py
```diff
@@ -0,0 +1,424 @@
...
+Holds the ROP-node callbacks and dialog wiring invoked from the Deadline Cloud
+HDA. Job-bundle generation is delegated to the headless :mod:`submitter` engine
+so the GUI and any headless consumer produce bundles from the same
+settings-driven contract.
+"""
+
+import os
+import sys
+import traceback
+from typing import Any, cast
```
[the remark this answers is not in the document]
leon-li-inspire: Addressed in `a7416f7`. The `Any` reference lived only inside a *quoted* cast annotation — `cast("list[dict[str, Any]]", queue_parameters)` — so CodeQL's AST scan saw the import as unused, while mypy/ruff still required it (the string is a forward reference). Rather than drop the import (which broke mypy + ruff `F401` the other way), I unquoted the cast to `cast(list[dict[str, Any]], queue_parameters)`, so `Any` is now a genuine runtime reference. CodeQL and ruff both pass on the current head (`675d6d7`).

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -873,3 +409,263 @@ def get_houdini_environments(init_data_attachment: dict[str, Any]) -> list[dict[
...
+        rop = self._get_rop_node()
+        if rop is None:
+            raise RuntimeError(f"Cannot find ROP node at path: {self._rop_node_path}")
+
+        # Return the typed AssetReferences per the BaseSubmitter contract; callers
+        # that need a job-bundle dict call .to_dict() at the serialization
+        # boundary (see _create_job_bundle).
+        return _get_evaluated_asset_references(rop)
+
+    def _get_rop_node(self):
```
leon-li-inspire: Do we need this method? If so, should this be renamed to `def get_rop_node`

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -873,3 +409,261 @@ def get_houdini_environments(init_data_attachment: dict[str, Any]) -> list[dict[
...
+                "value": houdini_settings.max_failed_tasks_count,
+            },
+            {"name": "deadline:maxRetriesPerTask", "value": houdini_settings.max_retries_per_task},
+            {"name": "HipFile", "value": _get_hip_file()},
+            *get_queue_parameter_values_as_openjd(rop),
+        ]
+
+        if houdini_settings.include_adaptor_wheels:
+            parameter_values.append(
+                {"name": "AdaptorWheels", "value": houdini_settings.adaptor_wheels}
```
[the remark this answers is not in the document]
leon-li-inspire: Good catch — fixed in 2c78f62. `get_parameter_values` now gates the `AdaptorWheels` value append on `os.path.exists(adaptor_wheels)`, the same guard `get_job_template` uses for the matching parameterDefinition + job environment. So a toggled-on-but-missing wheels path no longer produces a parameterValue without a parameterDefinition (which CreateJob would reject). Added two tests: `test_adaptor_wheels_value_emitted_when_path_exists` and `test_adaptor_wheels_value_omitted_when_path_missing`.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Removed in 2c78f62 — it was unnecessary. `pyproject.toml` already declares a mypy override for `hou.*` with `ignore_missing_imports = true` (\[\[tool.mypy.overrides\]\]), and every other module (`_assets.py`, `queue_parameters.py`, `gui_submitter.py`, …) imports `hou` with a plain `import hou`. mypy stays clean without the pragma, so this is now consistent with the rest of the package.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Yes, it's needed — `_get_rop_node` is the single place that resolves `self._rop_node_path` into a `hou.node`, used by `get_settings`, `get_job_template`, `get_parameter_values` and `get_asset_references` (and patched by the unit tests). I'd keep it private (leading underscore) rather than rename to `get_rop_node`: it's an internal helper, not part of the `BaseSubmitter` contract, and external callers set the path via the public `set_rop_node_path()` instead. Happy to expose a public accessor if you have a caller that needs the resolved node directly.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/gui_submitter.py
```diff
@@ -0,0 +1,439 @@
...
+            # resolution and asset-manager construction inside the dialog's
+            # submission worker (all read from the local Deadline Cloud config).
+            # The job bundle written above already carries the parameter values,
+            # so job_parameters stays the (currently empty) queue_parameters
+            # override.
+            job_progress_dialog.start_job_submission(
+                job_bundle_dir=job_bundle_dir,
+                job_parameters=cast(list[dict[str, Any]], queue_parameters),
+                submitter_name="houdini",
+                submitter_version=version,
```
moonyuet:
```suggestion
```
See: https://github.com/aws-deadline/deadline-cloud/blob/b8aaaa9df9275e7708a60613fe4706f21da996a2/src/deadline/client/ui/dialogs/submit_job_progress_dialog.py#L96-L99

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/gui_submitter.py
```diff
@@ -0,0 +1,439 @@
...
+            job_progress_dialog.start_submission(
+                farm_id,
+                queue_id,
+                storage_profile,
+                job_bundle_dir,
+                queue_parameters,
+                asset_manager,
+                deadline,
+                auto_accept=str2bool(get_setting("settings.auto_accept")),
+            )
```
moonyuet: Is the `start_submission` still used in other DCC submitters? Would it be more convenient to just implement back the `start_submission` as backward compatibility in initial deadline-cloud and let the overload method takes care of this. If there are arguments like farm_id, queue_id etc. it would just use `start_submission`. On the other hand, it would use `start_job_submission` if it is detected to have just bundle_dir, queue_parameters arguments.
This means that you dont need to mess around and implement this condition. And it would be helpful to re-direct the execution by looking how many and what type of arguments it has. 

Another approach I would think of is to make it forward compatible, we dont need to check whether there is `start_submission` anymore but rather `start_job_submission`, and we need to make sure the deadline-cloud dependencies installed at the latest version of deadline(if the PR is merged), and I believe this is easier for this case, as Houdini is already using hatch for installation.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/gui_submitter.py
```diff
@@ -0,0 +1,439 @@
...
+            # submission worker (all read from the local Deadline Cloud config).
+            # The job bundle written above already carries the parameter values,
+            # so job_parameters stays the (currently empty) queue_parameters
+            # override.
+            job_progress_dialog.start_job_submission(
+                job_bundle_dir=job_bundle_dir,
+                job_parameters=cast(list[dict[str, Any]], queue_parameters),
+                submitter_name="houdini",
+                submitter_version=version,
+            )
```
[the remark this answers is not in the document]
leon-li-inspire: Confirmed not a regression — the `auto_accept` setting is honored internally, so it doesn't need to be forwarded to `start_job_submission`. `start_job_submission` drives `deadline.client.api.create_job_from_job_bundle`, which reads `settings.auto_accept` from config directly (`_submit_job_bundle.py` lines ~527 and ~906) rather than taking it as an argument. That's different from the legacy positional `start_submission`, which required the caller to resolve/pass it — hence the kwarg only on that branch. So an auto-accept user is not prompted on the normal submit path. (The one case where a prompt still appears even with auto_accept on is the `from_gui` + *unknown asset paths* guard — deadline-cloud intentionally still confirms there; that's the known-asset-paths prompt, not an auto_accept regression.)

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -53,6 +55,160 @@ class RenderStrategy(Enum):
...
+
+        return settings
+
+    def get_job_template(
+        self,
+        settings: BaseSubmitterSettings,
+        host_requirements: Optional[dict[str, Any]] = None,
+    ) -> dict[str, Any]:
+        rop = self._require_rop()
+        return _get_job_template(rop, host_requirements)
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in 257d296. The build methods are now settings-driven: `get_job_template` overrides `name`/`description` from `settings`, and `get_parameter_values` builds the scalar `deadline:*` values (priority, targetTaskRunStatus, maxFailedTasksCount, maxRetriesPerTask) and adaptor-wheels from `settings` — so a consumer's edits after `get_settings()` are honored (matching deadline-cloud-for-maya #435). One deliberate exception: the **frame range** stays ROP-derived. A Houdini job's per-step frames/dependencies/wedges are produced by `hscript render` over the live ROP network and can't be reconstructed from a single `frame_list` string, so `settings.frame_list` is informational only (documented on `HoudiniSubmitterSettings`). The GUI path passes `get_settings()` into the engine, so node values still drive the bundle — verified byte-identical.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -53,6 +55,160 @@ class RenderStrategy(Enum):
...
+        # _get_parameter_values already emits the ROP's queue parameters
+        # (CondaPackages/CondaChannels via get_queue_parameter_values_as_openjd).
+        # Only append caller-supplied queue_parameters whose names are not
+        # already present, otherwise CreateJob rejects duplicate parameterValues.
+        existing_names = {param["name"] for param in parameter_values}
+        parameter_values.extend(
+            {"name": param["name"], "value": param["value"]}
+            for param in queue_parameters
+            if param["name"] not in existing_names
+        )
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in 257d296. Instead of silently keeping the node value (or the caller's), `get_parameter_values` now **raises `DeadlineOperationError` on any name overlap** between the job's own parameters and the caller-supplied `queue_parameters`, matching deadline-cloud-for-maya #435. The two sets are expected to be disjoint; a collision means they were never reconciled, so failing loudly is safer than dropping a value the consumer set. Disjoint `queue_parameters` are appended as before. Covered by `test_get_parameter_values_raises_on_queue_param_collision`.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -53,6 +56,223 @@ class RenderStrategy(Enum):
...
+
+        # Caller-supplied queue_parameters must be disjoint from the job's own
+        # parameters. On overlap we raise rather than silently pick a winner
+        # (matching deadline-cloud-for-maya) -- the two sets come from different
+        # tabs/sources and a collision means they were never reconciled, so a
+        # silent node-wins or caller-wins would drop a value the consumer set.
+        parameter_names = {param["name"] for param in parameter_values}
+        queue_parameter_names = {param["name"] for param in queue_parameters}
+        overlap = parameter_names & queue_parameter_names
+        if overlap:
```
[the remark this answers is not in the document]
leon-li-inspire: Great catch — you're right, this would have broken every real headless submission. Fixed in e8d4661: replaced the raise with **caller-wins de-dup**. The node emits its own queue parameters (CondaPackages/CondaChannels) via `get_queue_parameter_values_as_openjd`, and AYON passes the *same* queue parameters as resolved values, so they overlap by design — raising was wrong. Now the caller's value supersedes the node's on a name collision (mirroring the GUI's `set_queue_parameter_values_from_openjd`, which writes dialog edits onto the node before the bundle is read), de-duplicated so CreateJob sees no duplicates. (This is where Houdini differs from deadline-cloud-for-maya: Maya's job params are disjoint from queue params, so its raise-on-overlap is a safe sanity check; Houdini folds queue params into its own list, so caller-wins de-dup is the correct semantics.) New test uses the realistic CondaPackages-in-both case.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -53,6 +56,223 @@ class RenderStrategy(Enum):
...
+            {"name": "deadline:targetTaskRunStatus", "value": houdini_settings.initial_status},
+            {
+                "name": "deadline:maxFailedTasksCount",
+                "value": houdini_settings.max_failed_tasks_count,
+            },
+            {"name": "deadline:maxRetriesPerTask", "value": houdini_settings.max_retries_per_task},
+            {"name": "HipFile", "value": _get_hip_file()},
+            *get_queue_parameter_values_as_openjd(rop),
+        ]
+        if houdini_settings.include_adaptor_wheels and os.path.exists(
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in e8d4661. `get_job_template` now reconciles the AdaptorWheels `parameterDefinition` + `OverrideAdaptor` `jobEnvironment` with the same settings-driven condition `get_parameter_values` uses for the AdaptorWheels *value* (`include_adaptor_wheels and os.path.exists(adaptor_wheels)`), via a new `_apply_adaptor_wheels_override` helper: it adds the override when settings enable it and strips it otherwise, regardless of what the node-driven `_get_job_template` produced. So template and parameter values can no longer desync (definition-without-value or value-without-definition → CreateJob reject). It's idempotent, so the GUI path (settings read from the node) is unchanged. Added add/strip consistency tests.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter_api.py
```diff
@@ -0,0 +1,164 @@
...
+        # (via get_queue_parameter_values_as_openjd), which include
+        # CondaPackages/CondaChannels. Only append queue_parameters whose names
+        # are not already present, otherwise the job has duplicate
+        # parameterValues entries and CreateJob fails.
+        existing_names = {param["name"] for param in parameter_values}
+        parameter_values.extend(
+            {"name": param["name"], "value": param["value"]}
+            for param in queue_parameters
+            if param["name"] not in existing_names
+        )
```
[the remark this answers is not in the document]
leon-li-inspire: Outdated — this was on `submitter_api.py`, which no longer exists. The PR was re-scoped and rebased onto mainline after #372 merged; the engine now lives in `submitter.py` as `HoudiniSubmitter`. The queue-parameter precedence this comment raised is now handled by caller-wins de-dup in `get_parameter_values` (see the thread on submitter.py L254).

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/gui_submitter.py
leon-li-inspire: Outdated after the rebase onto mainline (post-#372): there's no `gui_submitter.py` anymore, and the `start_submission`/`start_job_submission` handling is now entirely in #372's merged GUI code (`submitter.py`), which this PR builds on rather than modifies. This PR only adds the headless `HoudiniSubmitter` engine, so the concern no longer applies here.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -567,13 +822,23 @@ def _create_job_bundle(
...
 ) -> None:
+    # Drive the unified HoudiniSubmitter engine so the GUI submit/export path
+    # and headless consumers (e.g. AYON) build the template + parameter values
+    # through one implementation. Pass the submitter's own get_settings() (read
+    # from this ROP) so the settings-driven builders see the node's real values;
+    # the produced bundle matches what the node configures.
+    # The dialog-provided asset_references are written as-is (they carry the
+    # user's Job Attachments edits) rather than re-scanned.
+    submitter = HoudiniSubmitter(rop_node.path())
+    settings = submitter.get_settings()
```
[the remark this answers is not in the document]
leon-li-inspire: Good catch — fixed in `083e703`.

`get_settings` now takes `scan_assets: bool = True` (default keeps base-contract callers like the AYON create/publish path unchanged), and `_create_job_bundle` (the GUI submit/export path) passes `scan_assets=False`. That path writes `asset_references.yaml` from the dialog-provided references and never reads `settings.input_*`/`output_directories`, so the scene + USD walk is skipped rather than run-and-discarded. Scalars and the frame range are still collected from the ROP. Added `test_get_settings_scan_assets_false_skips_scene_scan` asserting the scan isn't invoked while scalars are still populated.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -53,6 +55,258 @@ class RenderStrategy(Enum):
...
+                    settings.input_filenames = sorted(refs.input_filenames)
+                    settings.input_directories = sorted(refs.input_directories)
+                    settings.output_directories = sorted(refs.output_directories)
+                    if settings.output_directories:
+                        settings.output_path = settings.output_directories[0]
+                except Exception:
+                    # Scene scan can fail (e.g. missing/locked nodes); fall back
+                    # to the hip file / project path already set rather than
+                    # breaking settings collection.
+                    settings.input_filenames = [hip_file] if hip_file else []
```
kenyrish-amazon: Instead of repeating the same thing 3 times, could this be collapsed into one?

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -53,6 +55,258 @@ class RenderStrategy(Enum):
...
+
+    def set_rop_node_path(self, rop_node_path: str) -> None:
+        """Point the submitter at the Deadline Cloud ROP node to submit.
+
+        Lets external callers (e.g. the AYON create/publish plugins) seed the
+        node without reaching into the private attribute.
+        """
+        self._rop_node_path = rop_node_path
+
+    def _get_rop_node(self):
```
kenyrish-amazon: Nit: this doesn't specify a return type

kenyrish-amazon (dismissed):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Fixed in `20c23d9` — annotated as `def _get_rop_node(self) -> Optional["hou.Node"]:` (returns the resolved node, or `None` when no path is set). `mypy` clean.

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
leon-li-inspire: Collapsed in `20c23d9`. The three identical `settings.input_filenames = [hip_file] if hip_file else []` branches (no ROP / scan skipped / scan failed) are gone — the hip-file baseline is now set once right after computing `hip_file`, and the scene scan only *overwrites* it on success. Behavior is unchanged; the failed-scan `except` is now a bare `pass` keeping that baseline.

kenyrish-amazon (approved):

jairaws (approved):

src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/submitter.py
```diff
@@ -53,6 +55,258 @@ class RenderStrategy(Enum):
...
+            if scan_assets:
+                try:
+                    refs = _get_scene_asset_references(rop)
+                    settings.input_filenames = sorted(refs.input_filenames)
+                    settings.input_directories = sorted(refs.input_directories)
+                    settings.output_directories = sorted(refs.output_directories)
+                    if settings.output_directories:
+                        settings.output_path = settings.output_directories[0]
+                except Exception:
+                    # Scene scan can fail (e.g. missing/locked nodes); keep the
```
ttblanchard, line 194: Did we loose a log line here during rebasing? It looks like there was a previous comment that we should have one https://github.com/aws-deadline/deadline-cloud-for-houdini/pull/360#discussion_r3592335268

PR #364, opened 2026-07-08
feat: Run pre-GUI hooks in the Houdini render submitter

leon-li-inspire:

Studios want to pre-populate submitter dialog fields (job name, description, queue/job
parameters) from their own pipeline logic *before* the Houdini submitter dialog opens — the same
"pre-GUI hook" capability being added across the DCC submitters (Maya, Nuke). deadline-cloud
provides the Qt-free entry point `run_pre_gui_hooks` / `PreGuiHookContext` and the generic
`apply_pre_gui_output`; the Houdini submitter needs to call them before building its dialog.

This mirrors the equivalent Maya and Nuke submitter changes.

Call `run_pre_gui_hooks` in the submitter panel's `onCreateInterface`, just before constructing
`SubmitJobToDeadlineDialog`:

- Build `ui_settings` (`HoudiniSubmitterUISettings`) and an empty `shared_parameter_values` up
  front, and pass a copy of the parameters into the hook context.
- Houdini has no on-disk job bundle at pre-GUI time, so hooks are sourced from `DEADLINE_HOOKS_DIR`
  only (`bundle_dir=None`), gated by `settings.allow_environment_hooks`.
- The confirmation prompt is skipped when `settings.auto_accept` is set; otherwise the standard Qt
  confirmation dialog (`qt_hook_confirmation`) is shown, parented to `hou.qt.mainWindow()` to match
  the submitter's other dialogs (`submit_callback`, `login_callback`, `settings_callback`).
- The merged output is applied with deadline-cloud's **generic** `apply_pre_gui_output`.
  `HoudiniSubmitterUISettings` has no `.parameters` list (unlike the standalone submitter's
  `JobBundleSettings`), so `name` / `description` are written onto the settings object and every
  hook parameter flows to the shared values the dialog is seeded with. No DCC-owned mapping helper
  is needed — the core function handles this case directly.

Sets the `deadline` dependency floor to `>= 0.60.1, < 0.61`. The
`deadline.client.ui.pre_gui_hooks` module (`run_pre_gui_hooks` / `PreGuiHookContext` /
`apply_pre_gui_output`) first ships in **0.60.1** — the 0.60.0 release does **not** contain it
(the core change, aws-deadline/deadline-cloud#1255, landed after the 0.60.0 tag), so the floor
must be 0.60.1.

Also mocks `deadline.client.ui.deadline_authentication_status` in the unit-test bootstrap:
deadline-cloud >= 0.60 made `deadline.client.ui.dialogs` eagerly import
`submit_job_to_deadline_dialog`, which runs `DeadlineAuthenticationStatus.getInstance()` at import
time. Under the tests' mocked qtpy that raises, so it is mocked out (matching the Nuke submitter's
bootstrap). This is a pre-existing incompatibility surfaced by the version bump, not specific to
the pre-GUI path.

- **No behavior change when no hooks are configured.** With no `DEADLINE_HOOKS_DIR` env hooks (or
  with env hooks disabled), `run_pre_gui_hooks` returns an empty dict, `apply_pre_gui_output` is a
  no-op, and the dialog opens with the same empty shared values as before.
- When env hooks are present and enabled, they run (with a confirmation prompt unless
  `auto_accept` is set) and may pre-populate the job name, description, and parameters.

- `black --check`, `ruff check`, `mypy`: clean.
- New headless unit tests (`test/unit/deadline_submitter_for_houdini/test_pre_gui_hooks.py`), 4/4
  passing — they drive the **real** core `apply_pre_gui_output` against a real
  `HoudiniSubmitterUISettings`, guarding the DCC-owned contract: name/description application, hook
  parameters flowing to shared values (no `.parameters` list), empty-output no-op, and partial
  output.
- The **full unit suite (171 tests) passes against the released `deadline 0.60.1`** in a clean
  virtualenv with the package installed editable. CI (all Python legs across
  ubuntu/macos/windows × 3.9–3.13) is green.

- Have you run the unit tests? Yes.

Not run locally — the full submitter path requires a running Houdini, so it is validated in the
integration suite in CI.

No files were added or removed from `src/` (only `panel/submitter_panel.py` was modified);
`installer/` was not touched. Installer tests are unaffected.

Docstrings/comments were added at the new hook call site. No user-facing docs, README, or
adaptor init-data/run-data schemas are affected (this is a submitter-side, pre-dialog change).

No — the change is backward compatible and inert when no pre-GUI hooks are configured. The adaptor
interface (init-data / run-data) is unchanged. The required deadline-cloud release (`0.60.1`,
which ships `pre_gui_hooks`) is published, so there is no outstanding merge-order dependency.

----


src/deadline/houdini_submitter/panel/submitter_panel.py
```diff
@@ -11,8 +11,15 @@
...
 from qtpy.QtCore import Qt  # type: ignore
+from deadline.client.config import get_setting, str2bool
 from deadline.client.job_bundle.submission import AssetReferences
 from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import SubmitJobToDeadlineDialog
+from deadline.client.ui.pre_gui_hooks import (  # pylint: disable=import-error
+    PreGuiHookContext,
+    apply_pre_gui_output,
+    qt_hook_confirmation,
+    run_pre_gui_hooks,
+)
```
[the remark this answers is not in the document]
leon-li-inspire: Verified against the **published** `deadline 0.60.1` (the pinned floor is now `>= 0.60.1, < 0.61`):

- All four symbols import cleanly: `python -c "from deadline.client.ui.pre_gui_hooks import PreGuiHookContext, apply_pre_gui_output, qt_hook_confirmation, run_pre_gui_hooks"` succeeds. The `pre_gui_hooks` module ships in 0.60.1 (aws-deadline/deadline-cloud#1255); 0.60.0 predated it, which is why the floor is 0.60.1 rather than 0.60.0. The `pylint: disable=import-error` is only because the linter env may not have deadline-cloud installed — the module is genuinely present at the pinned range.
- Good catch on the related item: the stale `submitter_info` comment and its now-unnecessary `# type: ignore` have been removed in aea8e91 (`mypy` passes without it against 0.60.1).

The full unit suite (171 tests) passes against released 0.60.1 and CI is green, so the module-level imports resolve as expected.

src/deadline/houdini_submitter/panel/submitter_panel.py
```diff
@@ -11,8 +11,15 @@
     QVBoxLayout,
 )
 from qtpy.QtCore import Qt  # type: ignore
+from deadline.client.config import get_setting, str2bool
```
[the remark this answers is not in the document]
leon-li-inspire: Verified against the **published** `deadline 0.60.1` (the pinned floor): `str2bool` **is** re-exported from the `deadline.client.config` package `__init__` — it is listed in `deadline.client.config.__all__`, and `from deadline.client.config import get_setting, str2bool` succeeds. So this top-level import does not raise `ImportError` at the pinned range and `onCreateInterface` loads fine.

The `submitter.py` split you noticed (`str2bool` from `.config_file`, `get_setting` from the package) is just a stylistic inconsistency in that older file, not a signal that the package re-export is missing — both import paths are valid in 0.60.1. Happy to align `submitter.py` to the package-level import in a follow-up for consistency if preferred, but it is not required for correctness here.

test/unit/deadline_submitter_for_houdini/__init__.py
```diff
@@ -5,6 +5,11 @@
 
 # we must mock UI code
 mock_modules = [
+    # deadline-cloud >= 0.60 eagerly imports submit_job_to_deadline_dialog from
```
leon-li-inspire: Remove this comment as it is redundant

pyproject.toml
```diff
@@ -31,7 +31,10 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline >= 0.49,< 0.51",
+    # run_pre_gui_hooks / PreGuiHookContext / apply_pre_gui_output are required for the pre-GUI
```
leon-li-inspire: Remove this comment as this comment is redundant.

leon-li-inspire: Done — removed in efb942a.

test/unit/deadline_submitter_for_houdini/__init__.py
leon-li-inspire: Done — removed in efb942a.

jericht (approved):

crowecawcaw (approved):

